PR #5, opened 2016-03-15
update tag key consts and tags for new names

djspoons:


recorder.go
```diff
@@ -38,9 +38,11 @@ const (
 	ParentSpanGUIDKey = "parent_span_guid"
 
 	ComponentNameKey = "component_name"
-	ComponentGUIDKey = "component_guid"
+	GUIDKey          = "guid"
+	ComponentGUIDKey = "guid" // TODO remove legacy key
```
djspoons: I can delete the "legacy" ones now if you'd like.

PR #6, opened 2016-03-23
Adds API for explicit trace collection request

jmacd:
This change will break existing callers of lightstep.NewTracer due to a minor change in the Options struct.


jmacd:
Clearly this depends on the pending review in basictracer-go...

djspoons:
Sorry if I am sort of late to this party (or if there is some reason to Just Get This Done Now), but there are a couple of things I'd like to consider:
 - does BasicTracer need to be involved at all? If yes, then perhaps it's not the right abstraction. (Many implementation that use BasicTracer will not support late collection.) Or perhaps we should wrap BasicTracer to extend its functionality in a LightStep-specific way.
 - more high-level: should this be part of the tracer? perhaps the tracer should just let you extract the span id and then you use a different library to do the rest?
 - even more high-level: do we need to write a library for this at all? It's just a single HTTP request (there isn't any interesting buffering, retry, or error handling), right?

djspoons:
More concretely, why doesn't a client just write:
```go
http.Get("http://api.lightstep.com/v1/collect?span_guid=" + span.GUID())
```
or similar

bensigelman:
@djspoons re `http.Get`, there is also the auth issue to consider...

re whether we need to make this part of `basictracer`, there is a need to at least expose span details... but the service stub piece could be left out for now, [I think](https://github.com/opentracing/basictracer-go/pull/18/files#r57272967) (upon reflection, anyway).

jmacd:
(Updated for the ongoing OT/basictracer-go PR.)

The "auth issue" is addressed by the mechanism in this change regardless of the URL syntax, assuming we want to offer both auth-by-HTTP-header and auth-by-field-in-JSON-request modes.

This discussion left me pondering the state-of-the-art in RESTful API design. There is the attempt to standardize around resources and collections in the URL, so we might write

POST /api/span/<identifier>/assemble/trace
or
POST /api/span/<identifier>/action/assemble_trace

returning 202 on success (asynchronous request accepted).

djspoons:
On Auth: I would advocate for keeping it out of the body moving forward. Most examples I've seen keep auth separate from the data; many use HTTP basic auth or another header. Some quick examples: 
 - Basic auth:
https://docs.influxdata.com/influxdb/v0.11/administration/authentication_and_authorization/#authenticate-using-the-http-api
https://github.com/SumoLogic/sumo-api-doc/wiki/search-api
https://www.hostedgraphite.com/docs/#http-post
 - Non-basic-auth HTTP header:
https://support.stackdriver.com/customer/portal/articles/1491766-sending-custom-application-metrics-to-the-stackdriver-system
 - Query parameter:
https://github.com/DataDog/datadogpy/blob/032759e31909e83cbb24b0a155b15e5f0042190b/datadog/api/base.py#L74

---
On POST, URLs, etc. I really like the idea of standardizing around resources. Small nit: in this case I think the resource in question is a trace or better yet, a trace charter, not a span. That's what it is internally (for good reasons) and we should just expose that concept in some form. That would easily handle both seed-span-based charters, as well as end-user and other tag-based ones.

jmacd:
Auth: I did a survey a while back and found others for this list:

https://raygun.com/docs/deployments (JSON data field)
https://rollbar.com/docs/deploys_bash (Form fields)
https://help.airbrake.io/kb/airbrake-faq/deploy-tracking (resource in path)
http://support.hockeyapp.net/kb/api/api-versions (HTTP header)

POST, URLs:

I feel like scrutinizing the choice of "Charter" for this conceptual
entity. My impression is that there is a charter (grant of powers) to
organize collection of traces, but that a specific request for a trace or
exemplars would be more of a policy (course of action).

ᐧ

collect.go
```diff
@@ -0,0 +1,91 @@
...
+
+var (
+	errNotLightStepTracer = errors.New("not a LightStep tracer")
+	errSpanIsTooOld       = errors.New("span is too old to assemble")
+)
+
+type httpError string
+
+// AssembleTraceForSpan requests trace assembly given a span of
+// interest to the caller.
```
bensigelman: I guess we should mention that `span` must not be `Finish()`ed...

collect.go
```diff
@@ -0,0 +1,91 @@
...
+	nanosPerMicro = 1000
+
+	assembleTraceAPIPath = "/api/v1/trace/assemble"
+
+	// HTTP header for conveying the access token for web APIs
+	accessTokenHeader = "LightStep-Access-Token"
+)
+
+var (
+	errNotLightStepTracer = errors.New("not a LightStep tracer")
```
bensigelman: nit: should all errors be exported so that callers can equality-test them?

bensigelman:
Code LGTM, though for what it's worth I don't have an opinion about the REST-ish URLs and am thus leaving that in your capable hands :)

As for "Charter": regardless of whether it's the right name or not, there are charter types in the system that deal with single traces as well as charter types that handle all traces for a given user. The primitive to manage the literal assembly of a single trace is called a `TraceJoiner` (probably should have been `TraceAssembler` -- oh well?), btw.

jmacd:
I feel that this is not the time or place to (re-) design the web APIs. The question I would apply, especially concerning auth, is which technique is the most secure (i.e., least likely to leak API keys).

bensigelman:
@jmacd they'll all (headers, json, whatever) be sent over TLS – equivalent at that point, yes?

jmacd:
I suppose so.

jmacd:
Squashed.

djspoons:
On Mar 30, 2016 6:37 PM, "Joshua MacDonald" <notifications@github.com>
wrote:

>
> Auth: I did a survey a while back and found others for this list:
>
> https://raygun.com/docs/deployments (JSON data field)
> https://rollbar.com/docs/deploys_bash (Form fields)
> https://help.airbrake.io/kb/airbrake-faq/deploy-tracking (resource in
path)
> http://support.hockeyapp.net/kb/api/api-versions (HTTP header)

Thanks!

>
> POST, URLs:
>
> I feel like scrutinizing the choice of "Charter" for this conceptual
> entity. My impression is that there is a charter (grant of powers) to
> organize collection of traces, but that a specific request for a trace or
> exemplars would be more of a policy (course of action).

I don't have any feelings about the word itself, but I think it's important
to separate the trace from the thing-that-assembles-traces ("TTAT" or maybe
"assembler"?). The TTAT may assemble many traces over its lifetime, it may
be ephemeral unlike traces, which often live forever (e.g. a span-seeded
assembly only lives for a few minutes), it may fail for very different
reasons than just loading a trace (e.g. the collector horizon is not far
enough away), and it may have other status to convey (e.g. still assembling
for 30 more seconds).

Your charter-vs-policy point is well taken, but this probably ends up in
the middle somewhere (the TTAT might decide not to assemble a trace, and it
might have different ways to go about the assembly).

Looking at http://restcookbook.com/Resources/asynchroneous-operations/
this would argue for something that treated the request to create a one-off
assembler and the request to get that trace as almost the same, but offer
an additional channel to get more details, etc. for more advanced users.

Anyway, looking back at how much I've written, this probably deserves some
longer doc at this point, but just as an example of that sort of
interaction:


# >>> Force assembly of a trace using same method as reading it. (Use HEAD
to avoid returning the actual data.)
HEAD /api/trace?span_guid=12345678

# <<< If assembly hasn't started, return "Ok we're working on it!"
HTTP/1.1 202 Accepted
Location: /api/assembler/987654

# >>> That's all most users would care about but if they want to watch
progress, they'd ask for...
GET /api/assembler/987654
# <<< Which would return data about the assembler itself, e.g.
HTTP/1.1 200 OK
Content-type: application/json

{
 "assembly_complete": false,
 "assembly_end_time": 12318151922,
 "trace_url": "/api/trace?span_guid=12345678",
}

>On Wed, Mar 30, 2016 at 6:08 PM, spoons <notifications@github.com> wrote:
>
> > On Auth: I would advocate for keeping it out of the body moving forward.
> > Most examples I've seen keep auth separate from the data; many use HTTP
> > basic auth or another header. Some quick examples:
> >
> > - Basic auth:
>
> >
https://docs.influxdata.com/influxdb/v0.11/administration/authentication_and_authorization/#authenticate-using-the-http-api
> > https://github.com/SumoLogic/sumo-api-doc/wiki/search-api
> > https://www.hostedgraphite.com/docs/#http-post
> > - Non-basic-auth HTTP header:
> >
https://support.stackdriver.com/customer/portal/articles/1491766-sending-custom-application-metrics-to-the-stackdriver-system
> > - Query parameter:
> >
https://github.com/DataDog/datadogpy/blob/032759e31909e83cbb24b0a155b15e5f0042190b/datadog/api/base.py#L74
> >
> > ------------------------------
>
> >
> > On POST, URLs, etc. I really like the idea of standardizing around
> > resources. Small nit: in this case I think the resource in question is a
> > trace or better yet, a trace charter, not a span. That's what it is
> > internally (for good reasons) and we should just expose that concept in
> > some form. That would easily handle both seed-span-based charters, as
well
> > as end-user and other tag-based ones.
> >
> > —
> > You are receiving this because you authored the thread.
>
> > Reply to this email directly or view it on GitHub
> > <
https://github.com/lightstep/lightstep-tracer-go/pull/6#issuecomment-203701507
>
> >
>
> —
> You are receiving this because you were mentioned.
>
> Reply to this email directly or view it on GitHub
ᐧ

jmacd:
This is an ancient change to introduce access-token-based APIs so that self-requested "trace assembly" and other APIs can be based on access token, not username and password

bg451:
@jmacd sweet thank you, this is super useful.

PR #13, opened 2016-04-29
add makefile to generate thrift files

djspoons:
also, trivial changes to thrift generated files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: thrift
+
 # LightStep-specific: rebuilds the LightStep thrift protocol files.  Assumes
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,6 @@
+# LightStep-specific: rebuilds the LightStep thrift protocol files.  Assumes
```
bcronin: I think (technically) you also want to put a 

```make
.PHONY: thrift
```

somewhere before the rule so Make knows `thrift` is a pure target, not a file that's being generated.

djspoons: cool thanks

PR #14, opened 2016-05-11
Update LightStep Thrift generated code

bcronin:
## Summary

Updates the generated Thrift code to match the client library metrics updates.  

Note: this only updates the generated code and does not yet update the client library itself to use the metrics.


djspoons:
screen shot 2016-05-10 at 7 41 10 pm

Seems fine to me :)

PR #15, opened 2016-06-22
Add spans.dropped metric

bcronin:
Adds @sambati's dropped spans fix to the library.  

(On minor difference from the original diff is this also *no longer* reports the `Counters` field as well, as that data is superseded by the `InternalMetrics` data.)


bcronin:
FYI @sambati , @jmacd

jmacd:
LGTM

PR #20, opened 2016-07-05
Use atomic operations to track dropped spans; do not reset after a su…

jmacd:
…ccessful RPC (was a bug)

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -225,7 +225,7 @@ func (r *Recorder) RecordSpan(raw basictracer.RawSpan) {
 		return
 	}
 
-	atomic.AddInt64(&r.counters.droppedSpans, int64(r.addSpans.buffer([]basictracer.RawSpan{raw})))
+	atomic.AddInt64(&r.counters.droppedSpans, int64(r.buffer.addSpans([]basictracer.RawSpan{raw})))
 }
 
 func (r *Recorder) Flush() {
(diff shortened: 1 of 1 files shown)


recorder.go
```diff
@@ -224,7 +225,7 @@ func (r *Recorder) RecordSpan(raw basictracer.RawSpan) {
 		return
 	}
 
-	r.counters.droppedSpans += int64(r.buffer.addSpans([]basictracer.RawSpan{raw}))
+	atomic.AddInt64(&r.counters.droppedSpans, int64(r.addSpans.buffer([]basictracer.RawSpan{raw})))
```
jmacd: (nice typo)

bg451:
LGTM 
image

recorder.go
```diff
@@ -349,21 +353,22 @@ func (r *Recorder) Flush() {
...
+		atomic.AddInt64(&r.counters.droppedSpans, int64(r.buffer.addSpans(rawSpans))+droppedPending)
 		r.lock.Unlock()
 		return
 	}
 
 	// Reset the buffers
 	r.reportOldest = now
 	r.reportYoungest = now
-	// TODO: this ends up discarding counts coming in during the RPC
-	r.counters = counterSet{}
```
djspoons: This is definitely a bug, so great. I'm not convinced that the atomic int is needed (isn't it always read/written under the lock?) but happy to hear more.

jmacd: I'm leaving a TODO in my local copy to remove the atomics. Although we
could probably rewrite this to not take a lock (using a channel?) and then
the atomics would be handy.
ᐧ

PR #22, opened 2016-07-28
Add a lightstep tracer option to drop all span logs.

nimeshksingh:
It just sets the corresponding basictracer option.


djspoons:
LGTM

PR #28, opened 2016-08-31
use docker and sync

djspoons:
Hooray! Don't install gRPC/protobufs, just use Docker!


bensigelman:
🎉

Makefile
```diff
@@ -1,12 +1,19 @@
...
 
 proto:
-	@git submodule update --init
-	cd lightstep-tracer-common && protoc --go_out=plugins=grpc:. --go_out=../collectorpb/ collector.proto
-	@rm lightstep-tracer-common/collector.pb.go
+	@if [ ! -r lightstep-tracer-common/collector.proto ]; then \
+	  echo "Must run 'git submodule update --init' before generating protobuf stubs"; \
+	  false; \
+	fi
+	docker run -v $(shell pwd)/lightstep-tracer-common:/input:ro -v $(shell pwd)/collectorpb:/output \
```
bensigelman: I don't want to know how much pain is hidden behind this line :)

djspoons: Actually not that bad since I sort of copied the thrift version of this... I got more confused by running
```sh
git submodule update --init
```
on every make invocation, since that was changing git state (and reverting the submodule to a previous committed version).

bcronin:
:+100:

PR #32, opened 2016-09-06
Alice/grpc go

aliceadelef:
Enabling GRPC for the Go client


recorder.go
```diff
@@ -355,7 +420,8 @@ func (r *Recorder) Flush() {
 	r.reportInFlight = true
 	r.lock.Unlock() // unlock before making the RPC itself
 
-	resp, err := r.backend.Report(r.auth, req)
+	// Question: Where does context come in?
```
jmacd: context.Background() is appropriate here, since we're a background thread and there's no user activity tied to this Flush.

recorder.go
```diff
@@ -178,8 +190,8 @@ func NewRecorder(opts Options) basictracer.SpanRecorder {
 	if _, found := opts.Tags[ComponentNameKey]; !found {
 		opts.Tags[ComponentNameKey] = path.Base(os.Args[0])
 	}
-	if _, found := opts.Tags[GUIDKey]; !found {
-		opts.Tags[GUIDKey] = genSeededGUID()
+	if _, found := opts.Tags[GUIDKey]; found {
+		panic(fmt.Sprintf("Passing in your own %v is no longer supported", GUIDKey))
```
djspoons: Sorry if I missed a conversation, but my preference is to avoid all `panic`s in this library. Maybe just `maybeLogError`?

jmacd: Sorry, I missed that.
ᐧ

PR #37, opened 2016-09-08
Double buffer ls-tracer-golang spans

jmacd:
Also:

Encode Report (incl. serializing JSON payloads) outside the Recorder lock
Consolidate handling of default options
Return NoopTracer instead of Tracer with a nil Recorder (inside a non-nil interface, avoid panics)
Move oldest/youngest/dropped state into the buffer (simplifies things)
Document immutable vs mutable fields
Implement ReportTimeout option


recorder.go
```diff
@@ -420,28 +452,20 @@ func (r *Recorder) Flush() {
 		return
 	}
 
+	// not in flight => r.flushing has been reset
```
aliceadelef: not sure I follow this comment

recorder.go
```diff
@@ -188,20 +207,39 @@ type Recorder struct {
...
+
+	// Flush state.
+	reportInFlight    bool
+	lastReportAttempt time.Time
 
 	// We allow our remote peer to disable this instrumentation at any
 	// time, turning all potentially costly runtime operations into
 	// no-ops.
+	//
+	// TODO this should use atomic load/store to test disabled
```
bg451: CAS all the things

recorder.go
jmacd: Updated. Just to say that r.flushing must have been reset and is therefore ready to use.

recorder.go
jmacd: nah lets use double-checked locking.

spans_buffer.go
```diff
@@ -1,57 +1,74 @@
...
 }
 
-// addSpans returns the number of spans dropped (0 if all were added to the
-// buffer).
-func (b *spansBuffer) addSpans(spans []basictracer.RawSpan) (droppedSpans int) {
-	space := b.maxBufferSize - len(b.rawSpans)
-	count := space
-	if len(spans) < count {
-		count = len(spans)
+func (b *spansBuffer) mergeUnreported(a *spansBuffer) {
```
bg451: Really a pet peeve and not an actual issue: It took me an extra minute to comprehend what was going on since I have this idea that `a` should be the thing `b` is being merged into (if that makes sense).

jmacd: I had that happening to me too, but 'b' is for buffer. OK, I'll do something here.

jmacd: b->into
a->from

spans_buffer.go
```diff
@@ -1,57 +1,74 @@
...
+	"github.com/opentracing/basictracer-go"
+)
 
 type spansBuffer struct {
-	rawSpans      []basictracer.RawSpan
-	maxBufferSize int
+	rawSpans       []basictracer.RawSpan
+	dropped        int64
+	reportOldest   time.Time
+	reportYoungest time.Time
```
bg451: I'm on the bcronin boat of not being a fan of oldest/youngest. Perhaps `s/reportOldest/reportStart` and  `s/reportYoungest/reportEnd`?

jmacd: 👍

bg451:
A couple of comments, but other than that lgtm. I didn't see any glaring race conditions in the buffer swap code

jmacd:
Just realized I'm pushing updates to the wrong branch.
My Git/Github tolerance level is quite low.

bg451:
:git_parrot:

lightstep-tracer-common
```diff
@@ -1 +1 @@
-Subproject commit a2ac69205156c196ef5dbe7037fda7091580d1e9
+Subproject commit b1bdd09872f4e7d6439f613181997dedb1783458
```
djspoons: probably don't want this... maybe `git submodule update`?

jmacd: "git submodule update" doesn't do the right thing
but I did something similar, sure

ᐧ

jmacd: I checked-out master (in ls-tracer-common) and pulled. That gives me a different SHA so go figure.

recorder.go
```diff
@@ -188,20 +207,39 @@ type Recorder struct {
...
 	accessToken string
 
-	tracerID uint64
+	tracerID           uint64        // the LightStep tracer guid
+	verbose            bool          // whether to print verbose messages
+	maxReportingPeriod time.Duration // set by Options.MaxReportingPeriod
+	reportingTimeout   time.Duration // set by Options.ReportTimeout
+
+	// Remote service that will receive reports.
+	backend cpb.CollectorServiceClient
```
djspoons: I'm assuming this is thread-safe...?

recorder.go
```diff
@@ -188,20 +207,39 @@ type Recorder struct {
...
+	// prior to taking the lock, do please.
 	disabled bool
 }
 
-func NewRecorder(opts Options) basictracer.SpanRecorder {
+func NewRecorder(opts Options) *Recorder {
 	if len(opts.AccessToken) == 0 {
-		// TODO maybe return a no-op recorder instead?
-		panic("LightStep Recorder options.AccessToken must not be empty")
+		fmt.Println("LightStep Recorder options.AccessToken must not be empty")
```
djspoons: we also have `maybeLogInfo` for this sort of thing, but I suppose these one-off printfs are not going to spam anyone's terminal

recorder.go
```diff
@@ -188,20 +207,39 @@ type Recorder struct {
...
+		return nil // Note: this is a non-nil interface w/ a nil Recorder
```
djspoons: Somehow I'm not comforted by that comment. We can return `nil` here, but then I think we need to handle that case in the methods, e.g.
```go
func (r *Recorder) RecordSpan(raw basictracer.RawSpan) {
  if r == nil {
    return
  }
 ...
```

recorder.go
jmacd: I think this condition is serious enough that it deserves a message even without Verbose enabled.

recorder.go
jmacd: That's my understanding, but it's used only from the flush thread either way.

lightstep-tracer-common
djspoons: checkout-master-and-pull is very different than what I was suggesting.

`git submodule update` should bring the submodule (`lightstep-tracer-common`) back into sync with what the parent repo says about it--it's sort of the analogue of `git checkout` on an ordinary file. 

checkout-master-in-the-submodule-and-pull is like editing a normal file. If you do that you should also run `make proto` to keep the generated file in sync.

(OMG submodules)

spans_buffer.go
```diff
@@ -1,57 +1,74 @@
...
 }
 
-// addSpans returns the number of spans dropped (0 if all were added to the
-// buffer).
-func (b *spansBuffer) addSpans(spans []basictracer.RawSpan) (droppedSpans int) {
-	space := b.maxBufferSize - len(b.rawSpans)
-	count := space
-	if len(spans) < count {
-		count = len(spans)
+func (into *spansBuffer) mergeUnreported(from *spansBuffer) {
```
djspoons: nit: I might call it `mergeFrom`.

Also, a comment would be nice.

recorder.go
```diff
@@ -420,28 +455,21 @@ func (r *Recorder) Flush() {
 		return
 	}
 
+	// There is not an in-flight report, therefore r.flushing has been reset and
+	// is ready to re-use.
 	now := time.Now()
+	tmp := r.buffer
```
djspoons: why not this? ;)
```go
r.buffer, r.flushing = r.flushing, r.buffer
```

spans_buffer.go
```diff
@@ -1,57 +1,74 @@
...
+	"time"
 
-const defaultMaxSpans = 1000
+	"github.com/opentracing/basictracer-go"
+)
 
 type spansBuffer struct {
-	rawSpans      []basictracer.RawSpan
-	maxBufferSize int
+	rawSpans    []basictracer.RawSpan
```
djspoons: Total nit, but since you are here, I might call this `report` (or `partialReport`) rather than `spansBuffer`.

djspoons:
A couple of nits (and a submodule thing) but otherwise, looks great!

spans_buffer.go
jmacd: So I had it named mergeFrom and decided mergeUnreported would be more descriptive. Anyway. So.

spans_buffer.go
jmacd: I propose `reportBuffer`

jmacd:
OMG submodules I just updated the generated proto code instead of reverting the submodule update.

PR #38, opened 2016-09-09
enable span logs for grpc

aliceadelef:


recorder.go
```diff
@@ -313,29 +315,71 @@ func translateDurationFromOldesYoungest(ot time.Time, yt time.Time) uint64 {
...
 	return kvs
 }
 
-// TODO: Implement once OT logs have been updated
-func translateLogs(log []ot.LogData) []*cpb.Log {
-	return nil
+func translateEventAndPayload(e string, pl interface{}) []*cpb.KeyValue {
+	var kvs []*cpb.KeyValue
+	kvs = append(kvs, &cpb.KeyValue{Key: messageKey, Value: &cpb.KeyValue_StringValue{e}})
+	return translatePayload(pl, kvs)
```
jmacd: The above three lines could be written 
return translatePayload(pl, []*cpb.KeyValue{&cpb.KeyValue{Key: messageKey, Value: &cpb.KeyValue_StringValue{e}}})

(now that I read it, not sure it's better)

recorder.go
```diff
@@ -313,29 +315,71 @@ func translateDurationFromOldesYoungest(ot time.Time, yt time.Time) uint64 {
...
+	switch p := pl.(type) {
+	case map[string]interface{}:
+		for k, v := range p {
+			kvs = append(kvs, convertToKeyValue(k, v))
+			glog.Infof("the map is looks like {%v : %v}", k, v)
 		}
-		kvs = append(kvs, &kv)
+	case []interface{}:
+		for _, v := range p {
+			kvs = append(kvs, convertToKeyValue(payloadKey, v))
```
jmacd: I wonder if we've specified what happens for duplicate keys in these KV logs. probably haven't.

jmacd:
👍

recorder.go
aliceadelef: huh yeah... not sure. I'll add a note about that at least

PR #39, opened 2016-09-09
Alice/simplify logs for grpc

aliceadelef:


recorder.go
```diff
@@ -377,29 +378,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
...
 	var kvs []*cpb.KeyValue
 	kvs = append(kvs, &cpb.KeyValue{Key: messageKey, Value: &cpb.KeyValue_StringValue{e}})
-	return translatePayload(pl, kvs)
+	if pl == nil {
+		return kvs
+	}
+	jpl, err := json.Marshal(pl)
+	glog.Infof("the log is %v", string(jpl))
+	if err != nil {
+		return append(kvs, &cpb.KeyValue{Key: payloadKey, Value: &cpb.KeyValue_StringValue{fmt.Sprintf("%v", err)}})
```
aliceadelef: especially curious what you think about this

recorder.go
```diff
@@ -377,29 +378,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
...
-
 func translateEventAndPayload(e string, pl interface{}) []*cpb.KeyValue {
 	var kvs []*cpb.KeyValue
 	kvs = append(kvs, &cpb.KeyValue{Key: messageKey, Value: &cpb.KeyValue_StringValue{e}})
-	return translatePayload(pl, kvs)
+	if pl == nil {
+		return kvs
+	}
+	jpl, err := json.Marshal(pl)
+	glog.Infof("the log is %v", string(jpl))
```
djspoons: I think @jmacd will say something about glog

recorder.go
djspoons: Ideally the error would go in InternalMetrics.Logs (keep our errors out of users' logs)

recorder.go
```diff
@@ -377,29 +378,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
...
-		return nil
-	default:
-		kvs = append(kvs, convertToKeyValue(payloadKey, p))
-	}
-	return kvs
-}
-
 func translateEventAndPayload(e string, pl interface{}) []*cpb.KeyValue {
 	var kvs []*cpb.KeyValue
 	kvs = append(kvs, &cpb.KeyValue{Key: messageKey, Value: &cpb.KeyValue_StringValue{e}})
```
jmacd: Nit: I'd probably write this without an append, e.g., a one-liner:

kvs := []*cpb.KeyValue{&cpb.KeyValue{Key: ..., Value: ...}}

recorder.go
djspoons: Also, I think `fmt.Sprintf("%v", X)` == `fmt.Sprint(X)`

recorder.go
```diff
@@ -377,29 +378,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
...
-	return translatePayload(pl, kvs)
+	if pl == nil {
```
djspoons: The two lines above can just be 
```go
kvs := []*cpb.KeyValue{&cpb.KeyValue{Key: messageKey, Value: &cpb.KeyValue_StringValue{e}})}
```

recorder.go
jmacd: Good approach.
You can replace fmt.Sprintf("%v", err) with err.Error()

jmacd:
👍

djspoons:
Yeah, this is what I was thinking.

recorder.go
aliceadelef: @djspoons, @jmacd and I were thinking that since this error comes from marshaling user sent data, it might make sense for them to be aware of that. But I don't think we are too tied to the idea if you disagree.

recorder.go
aliceadelef: oh oops that was left over from debugging :|

PR #40, opened 2016-09-12
added reflection for type checking in translation to proto

aliceadelef:


recorder.go
```diff
@@ -371,9 +371,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
 	case bool:
 		kv.Value = &cpb.KeyValue_BoolValue{v}
 	default:
-		glog.Infof("value: %v, %T, is an unsupported type, and has been converted to string", v, v)
-		// TODO: use reflection so that not all custom types have to be converted to string
-		kv.Value = &cpb.KeyValue_StringValue{fmt.Sprint(v)}
+		r := reflect.ValueOf(v)
+		r = reflect.Indirect(r)
+		if r.Type().ConvertibleTo(reflect.TypeOf(float64(0))) {
```
djspoons: maybe worth calling these `TypeOf`s once? i.e., outside of any function
```go
var floatType reflect.Type = reflect.TypeOf(float(0))
```

recorder.go
```diff
@@ -371,9 +371,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
...
+			kv.Value = &cpb.KeyValue_DoubleValue{r.Convert(reflect.TypeOf(float64(0))).Float()}
+		} else if r.Type().ConvertibleTo(reflect.TypeOf(true)) {
+			kv.Value = &cpb.KeyValue_BoolValue{r.Bool()}
+		} else if r.Type().ConvertibleTo(reflect.TypeOf("")) {
+			kv.Value = &cpb.KeyValue_StringValue{r.String()}
+		} else {
+			glog.Infof("value: %v, %T, is an unsupported type, and has been converted to string", v, v)
```
djspoons: I think I would prefer `maybeLogInfo` here since 1. that way we can avoid glog and 2. it's not the worst thing if this happens and it'd by nice not to spam stderr.

recorder.go
```diff
@@ -371,9 +371,18 @@ func convertToKeyValue(k string, value interface{}) *cpb.KeyValue {
 	case bool:
 		kv.Value = &cpb.KeyValue_BoolValue{v}
 	default:
-		glog.Infof("value: %v, %T, is an unsupported type, and has been converted to string", v, v)
-		// TODO: use reflection so that not all custom types have to be converted to string
-		kv.Value = &cpb.KeyValue_StringValue{fmt.Sprint(v)}
+		r := reflect.ValueOf(v)
+		r = reflect.Indirect(r)
```
djspoons: I think this will be fine, but can you appease my neuroses by adding tests for 
```go
convertToKeyValue("k", nil)
```
and
```go
var p *int
convertToKeyValue("k", p)
```

djspoons:
Great! 🚢

jmacd:
👍

PR #41, opened 2016-09-14
moved error log to internal metrics logs

aliceadelef:


aliceadelef:
Not sure if pointer was the way to go... thoughts?

jmacd:
I'd say what you have now is more idiomatic in Go.

jmacd:
lgtm

PR #42, opened 2016-09-15
Reconnect clients every 5-6 minutes

jmacd:
Reconnect the client every 5-6 minutes
Allow use of insecure gRPC via Plaintext option
Add Recorder.Close method

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -344,16 +344,15 @@ func (r *Recorder) reconnectClient(now time.Time) {
 	if err != nil {
 		r.maybeLogInfof("could not reconnect client")
 	} else {
-		toClose := r.conn
-
 		r.lock.Lock()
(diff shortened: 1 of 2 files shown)


recorder.go
```diff
@@ -588,19 +644,28 @@ func (r *Recorder) shouldFlush() bool {
...
-		if r.shouldFlush() {
-			r.Flush()
+			if disabled {
+				return
+			}
+			if reconnect {
+				r.reconnectClient(now)
+			}
+			if shouldFlush {
+				r.Flush()
```
aliceadelef: not sure if this would actually be an issue, but would it make sense to flush and then reconnectClient? That way  in case the reconnect doesn't work at least you have flush what you currently have?

PR #44, opened 2016-09-18
Move to OpenTracing KeyValue logging

bensigelman:
cc @jmacd and @aliceadelef regarding the treatment of the reportBuffer, especially (which led to some refactoring). I decided to go for an encoding error count rather than internal log records and also changed the way parameters are passed around.

I cleaned up a few misc TODOs along the way, too.

### DIFF
--- a/logencoder.go
+++ b/logencoder.go
@@ -14,11 +14,17 @@ const (
 // An implementation of the log.Encoder interface
 type logFieldEncoder struct {
 	recorder        *Recorder
+	buffer          *reportBuffer
 	currentKeyValue *cpb.KeyValue
 }
 
-func marshalFields(recorder *Recorder, protoLog *cpb.Log, fields []log.Field) {
-	lfe := logFieldEncoder{recorder, nil}
+func marshalFields(
+	recorder *Recorder,
+	protoLog *cpb.Log,
+	fields []log.Field,
+	buffer *reportBuffer,
+) {
+	lfe := logFieldEncoder{recorder, buffer, nil}
 	protoLog.Keyvalues = make([]*cpb.KeyValue, len(fields))
 	for i, f := range fields {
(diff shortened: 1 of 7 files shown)


bensigelman:
(See PR 108 in the OT-Go repo for context)

logencoder.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"encoding/json"
+
+	cpb "github.com/lightstep/lightstep-tracer-go/collectorpb"
+	"github.com/opentracing/opentracing-go/log"
+)
+
+const (
+	ellipsis = "…"
```
aliceadelef: ? why "&"

thrift_rpc/logencoder.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"github.com/lightstep/lightstep-tracer-go/lightstep_thrift"
+	"github.com/lightstep/lightstep-tracer-go/thrift_0_9_2/lib/go/thrift"
+	"github.com/opentracing/opentracing-go/log"
+)
+
+const (
+	deprecatedFieldKeyEvent   = "event"
+	deprecatedFieldKeyPayload = "payload"
```
djspoons: You can delete the original versions of these (though the "event" one is called "message")

recorder.go
```diff
@@ -28,8 +26,9 @@ import (
 )
 
 const (
-	spansDropped  = "spans.dropped"
-	collectorPath = "/_rpc/v1/reports/binary"
+	spansDropped     = "spans.dropped"
+	logEncoderErrors = "log_encoder.errors"
+	collectorPath    = "/_rpc/v1/reports/binary"
```
djspoons: While you're here, `collectorPath` is unused.

bensigelman: (not unused AFAICT...)

PR #47, opened 2016-09-29
Add MaxLogsPerSpan option

### LINKED ISSUE #46: limit number of log events
We had an issue where a bug was causing endless retries and the memory inside basictracer grew indefinitely. It would be great if there was an option to only keep the last `X` log events inside a span.

Assuming we implement this functionality in basictracer, would it be ok with lightstep to pass along that option? (similar to how we pass `DropSpanLogs`)

RaduBerinde:
Basictracer now supports a `MaxLogPerSpan` options; plumbing this option in
Lightstep as well.

Fixes #46.

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -45,6 +45,7 @@ const (
 	defaultReportTimeout  = 30 * time.Second
 	defaultMaxLogKeyLen   = 256
 	defaultMaxLogValueLen = 1024
+	defaultMaxLogsPerSpan = 500
 
 	// ParentSpanGUIDKey is the tag key used to record the relationship
 	// between child and parent spans.
@@ -117,6 +118,9 @@ type Options struct {
 	// variable-length value types (strings, interface{}, etc).
 	MaxLogValueLen int `yaml:"max_log_value_len"`
 
+	// MaxLogsPerSpan limits the number of logs in a single span.
+	MaxLogsPerSpan int `yaml:"max_logs_per_span"`
+
 	// ReportingPeriod is the maximum duration of time between sending spans
 	// to a collector.  If zero, the default will be used.
 	ReportingPeriod time.Duration `yaml:"reporting_period"`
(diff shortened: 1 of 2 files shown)


recorder.go
```diff
@@ -117,6 +118,9 @@ type Options struct {
 	// variable-length value types (strings, interface{}, etc).
 	MaxLogValueLen int `yaml:"max_log_value_len"`
 
+	// MaxLogsPerSpan limits the number of logs in a single span.
```
bensigelman: remind me -- if someone sets this to `0`, it would actually inherit the basictracer default, correct?

RaduBerinde: No, we are setting `options.MaxLogsPerSpan` unconditionally after getting the `DefaultOptions()`. If we want that we can enclsoe that in a `if MaxLogsPerSpan != 0`

RaduBerinde:
It looks like the CI build is using an old `basictracer-go` (which doesn't build with the latest `opentracing-go`).

resospoons:
Yes, that happened with the previous build as well. I'll take a look.

resospoons:
The CircleCI cache needed to be flushed. Fixed!

bensigelman:
Sorry I forgot about this...

On one hand, I'm not crazy about the way the second set of defaults is expressed here. On the other hand, I don't want to make you change that since it would affect BasicTracer as well. So we can merge this and potentially clean up the options stuff separately.

RaduBerinde:
Thanks @bensigelman 

I can do some cleanup, what do you have in mind? Exporting the default value from basictracer as a constant that we use here as well?

bensigelman:
@RaduBerinde there's a lightstep-internal effort to canonicalize the semantics of tracer tuning options across platforms... until that's settled down, it's probably not worth the churn. (Aside from anything else, I'd rather do a "functional options" idiom for the initialization)

PR #49, opened 2016-10-07
cancel funcs should always be called on cancellable contexts

aybabtme:
Ohai!  I noticed this `context.CancelFunc` wasn't being called. Since the lifetime of the context is the RPC call happening right after it, I think a simple defer works fine.

> Calling the CancelFunc cancels the child and its children, removes the parent's reference to the child, and stops any associated timers. Failing to call the CancelFunc leaks the child and its children until the parent is canceled or the timer fires. 
> - https://golang.org/pkg/context

> Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete: [...]
> - https://golang.org/pkg/context/#WithTimeout

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -586,7 +586,8 @@ func (r *Recorder) Flush() {
 	r.lastReportAttempt = now
 	r.lock.Unlock()
 
-	ctx, _ := context.WithTimeout(context.Background(), r.reportingTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), r.reportingTimeout)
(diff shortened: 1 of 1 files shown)


jmacd:
My read of the context package documentation was that WithDeadline and WithTimeout will have their cancel functions called automatically, since they are based on time. Since we call a blocking RPC immediately after, the RPC should either finish or cancel itself, therefore I don't see the cancellation happening any sooner this way, either.

That said, I see no reason _not_ to apply your change.

PR #50, opened 2016-10-07
apply default options to all exported entry points

aybabtme:
Both `NewTracer` and `NewRecorder` accept an `Options` as argument, but only the `NewTracer` call will set proper defaults to the options. To make both entry points behave similarly, I think it makes sense to extract the defaults code into a method/function that then get used consistently in both `NewTracer` and `NewRecorder`.  

e.g: using a bare `Option`: 

```go
NewRecorder(Option{AccessToken: "..."})
``` 
will currently lead to a misbehaving recorder that has timeouts of 0s.

I don't think this is the most elegant way to go about this, I'm happy to rework it! Just want to bring the idea up.

Note: the diff doesn't show the extracted code, it seems to think I moved the first 3 lines of the `NewTracer` function.

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -138,12 +138,7 @@ type Options struct {
 	ReconnectPeriod time.Duration `yaml:"reconnect_period"`
 }
 
-// NewTracer returns a new Tracer that reports spans to a LightStep
-// collector.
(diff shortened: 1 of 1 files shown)


jmacd:
Looks good.

PR #54, opened 2016-10-22
use separate default host when UseGPRC is set

RaduBerinde:
Tested cockroachdb with and without `UseGPRC` set.
CC @jmacd


tamird:
:+1:

jmacd:
@djspoons I see this as a stop-gap until we get everything on the same host/port. OK with you?

djspoons:
LGTM and thanks

Issue #57, opened 2016-11-04
collectorpb no longer builds due to breaking change in grpc


peterbourgon:
```
ugh ~/src/github.com/lightstep/lightstep-tracer-go (master) git rev-parse --short HEAD
91ae78c
ugh ~/src/github.com/lightstep/lightstep-tracer-go (master) go build ./...
# github.com/lightstep/lightstep-tracer-go/collectorpb
collectorpb/collector.pb.go:631: undefined: grpc.SupportPackageIsVersion3
```

djspoons:
I vendored a (working) version of gRPC in https://github.com/lightstep/lightstep-tracer-go/pull/55

Let me know if that doesn't help

peterbourgon:
@djspoons [Libraries cannot vendor their deps](http://peter.bourgon.org/go-best-practices-2016/#dependency-management), so no, this doesn't help. Unfortunately.

djspoons:
For the record, no gGPC types escape through the public interface of lightstep-tracer-go.

I will take a closer look in the morning.

peterbourgon:
If it helps, you can check out the build failures I'm (still) getting over at https://github.com/go-kit/kit/pull/387

peterbourgon:
Thanks, that almost did it :) Can you please update your vendored copy of x/net/context? The currently vendored copy uses alias syntax `=>` in Go 1.8, which has been removed and no longer compiles on the current tip.

jmacd:
We this this addresses the issue:
https://github.com/lightstep/lightstep-tracer-go/pull/61

tedsuo:
I am assuming this issue is resolved, feel free to reopen this issue or start a new one if there is further work to be done.

PR #58, opened 2016-11-04
update for latest version of gRPC

djspoons:
besides for a makefile, this is just generated and vendored code


djspoons:
Fixes #57

PR #60, opened 2016-11-07
Remove vendor directory

tamird:
Vendoring in Go is not generally safe for libraries to use.

In this particular case, vendoring grpc results in vendoring
x/net/trace, which has a non-idempotent `init()` function, which panics
on invocations after the first. The result is that importers of this
package which also import x/net/trace will panic on init.

@djspoons


djspoons:
Given the frequency with which gRPC makes breaking changes, it pains me a little that we can't vendor it. (Though acknowledged that we shouldn't include x/net/trace for the reason you state.)

jmacd:
How about this:
https://github.com/lightstep/lightstep-tracer-go/pull/61

tamird:
#61 seems to unblock us for the moment, but I would still insist that vendoring in a library is fragile and that you guys should avoid it.

jmacd:
We're coming around to sharing this point of view, but we also feel that
not vendoring gRPC is fragile and want to isolate our customers from broken
builds.
ᐧ

djspoons:
Done in https://github.com/lightstep/lightstep-tracer-go/pull/70

PR #70, opened 2017-01-08
update makefile; remove vendor directory

djspoons:
- Makefile now points to new (but still private) location of the .thrift file
  - Also fix a sign error where the .thrift output might not be regenerated
- Remove the vendored copy of gRPC


Makefile
```diff
@@ -6,16 +6,16 @@ default: build
...
-ifeq (,$(wildcard $(LIGHTSTEP_HOME)/go/src/crouton/crouton.thrift))
-# LightStep-specific: rebuilds the LightStep thrift protocol files.  Assumes
-# the command is run within the LightStep development environment (i.e. the
-# LIGHTSTEP_HOME environment variable is set).
-lightstep_thrift/constants.go: $(LIGHTSTEP_HOME)/go/src/crouton/crouton.thrift
-	docker run -v "$(LIGHTSTEP_HOME)/go/src/crouton:/data" -v "$(PWD):/out" --rm thrift:0.9.2 \
-		thrift --gen go:package_prefix='github.com/lightstep/lightstep-tracer-go/',thrift_import='github.com/lightstep/lightstep-tracer-go/thrift_0_9_2/lib/go/thrift' -out /out /data/crouton.thrift
-	rm -rf lightstep_thrift/reporting_service-remote
-else
+ifeq (,$(wildcard $(GOPATH)/src/github.com/lightstep/common-go/crouton.thrift))
```
bensigelman: so this `ifeq` is basically an `if the thing to the right of the comma is not found`?

djspoons: Yes. Happy to do it other ways. Quick thinking/googling yields this as an alternative:
```
ifeq ($(shell test -e $(GOPATH)/src/github.com/lightstep/common-go/crouton.thrift && echo -n found),found)
...
```

PR #71, opened 2017-03-14
Implements all log.Encoder methods using EmitString

bg451:


thrift_rpc/logencoder.go
```diff
@@ -48,13 +48,28 @@ func (lfe *logFieldEncoder) EmitObject(key string, value interface{}) {
 	}
 }
 
-// All other log.Encoder methods do nothing in the thrift_rpc implementation.
```
bhs: kind of shameful :-/

Thanks for changing this. I guess the left side was written at a time when we assumed thrift was not long for this world.

PR #72, opened 2017-03-21
Encode all KV pairs instead of those that have the 'event' key

bg451:


tedsuo:
Thank god!

thrift_rpc/logencoder.go
```diff
@@ -25,27 +24,30 @@ type logFieldEncoder struct {
...
-			thriftPayload = string(jsonString)
-		}
-		if len(thriftPayload) > lfe.recorder.maxLogMessageLen {
-			thriftPayload = thriftPayload[:(lfe.recorder.maxLogMessageLen-1)] + ellipsis
-		}
-		lfe.logRecord.PayloadJson = thrift.StringPtr(thriftPayload)
+	var thriftPayload string
+	jsonString, err := json.Marshal(value)
+	if err != nil {
+		thriftPayload = fmt.Sprintf("Error encoding payload object: %v", err)
```
bcronin: I realize this PR is not introducing this, but since it's relevant to the affected code, it strikes as a "bad idea" that the same field _could_ contain a valid JSON string, an error string, or an un-parseable truncated JSON string with no hint to the receiver as to which it is. I thought we moved away from this a while back in the client libraries.

PR #73, opened 2017-04-20
Support lightstep.BinaryCarrier

jmacd:


basictracer/propagation_ls.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+	if bcarrier == nil {
+		*scarrier = base64.StdEncoding.EncodeToString(data)
+	} else {
+		*bcarrier = make([]byte, base64.StdEncoding.EncodedLen(len(data)))
+		base64.StdEncoding.Encode(*bcarrier, data)
+	}
+	return nil
+}
```
tedsuo: Small readability/optimization improvement: please consider this version of `Inject` that has fewer conditionals and extra allocations (and something similar for `Extract`).
```go
{
	sc, ok := spanContext.(SpanContext)
	if !ok {
		return opentracing.ErrInvalidSpanContext
	}

	data, err := proto.Marshal(&lightstep.BinaryCarrier{
		BasicCtx: &lightstep.BasicTracerCarrier{
			TraceId:      sc.TraceID,
			SpanId:       sc.SpanID,
			Sampled:      true,
			BaggageItems: sc.Baggage,
		}
	})
	if err != nil {
		return err
	}

	switch carrier := opaqueCarrier.(type) {
	case *string:
		*carrier = base64.StdEncoding.EncodeToString(data)
	case *[]byte:
		*carrier = make([]byte, base64.StdEncoding.EncodedLen(len(data)))
		base64.StdEncoding.Encode(*carrier, data)
	default:
		return opentracing.ErrInvalidCarrier
	}

	return nil
}
```

basictracer/propagation_ls.go
```diff
@@ -0,0 +1,105 @@
...
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+type lightstepBinaryPropagator struct{}
+
+type lightstepBinaryCarrier struct{}
+
+var BinaryCarrier lightstepBinaryCarrier
+
+func (_ *lightstepBinaryPropagator) Inject(
```
jmacd: Done

basictracer/propagation_ls.go
jmacd: Done

jmacd:
I added a TODO about consolidating packages.

RaduBerinde:
Hi @jmacd - quick question, what is the main motivation for the binary carrier? Is it expected to have less overhead than a `HTTPHeaders` carrier?  I'm trying to decide which one to use (I'm leaning towards `HTTPHeaders` so I can use the grpc-opentracing interceptors unmodified).

jmacd:
Yes, the BinaryCarrier is partly about efficiency, but also provides a path
for forward/backward compatibility in case a Tracer implementation uses an
alternative representation for SpanContext.

I believe the grpc-opentracing plugin should be changed to recognize binary
carriers, actually, but that will take some carefully worded arguments and
may not happen soon. (It came up today in another context, so it's worth
thinking about.)

ᐧ

RaduBerinde:
Thanks! Makes sense!

PR #77, opened 2017-04-26
Vendor the 32-bit signed timestamp offset fix

jmacd:


jmacd:
Reviewed / tested in person.

PR #80, opened 2017-05-16
Support explicit context IDs via LS-specific StartSpanOptions LS-948

jmacd:


basictracer/tracer.go
```diff
@@ -70,43 +85,54 @@ func (t *tracerImpl) StartSpan(
 	operationName string,
 	opts ...opentracing.StartSpanOption,
 ) opentracing.Span {
-	sso := opentracing.StartSpanOptions{}
+	sso := StartSpanOptions{}
 	for _, o := range opts {
-		o.Apply(&sso)
+		if lso, ok := o.(LightStepStartSpanOption); ok {
```
tedsuo: A type switch is a little more standard:

```go
switch o := o.(type) {
	case LightStepStartSpanOption:
		o.ApplyLS(&sso)
	default:
		o.Apply(&sso.Options)
}
```

jmacd: Ack. Thanks.

Issue #82, opened 2017-06-01
API for cleaning up a tracer


RaduBerinde:
In CockroachDB, we want to be able to start and stop using Lighstep, or change the Lightstep token (and potentially the collector address) without restarting. We can easily create a new tracer when the settings change; however, there is no API to clean up the background task of the previous tracer.

I know this is not part of the opentracing interface, so it would have to be a `Stop(opentracing.Tracer)` function that downcasts to the internal type.

djspoons:
@jmacd @tedsuo would the two of you take a look at this? Thanks

Issue #56, opened 2016-11-02
expose a method to close and drain the reporting loop


aybabtme:
I'd like to have a method to tell the tracer "no more spans are coming, please try to send all the spans you have buffered".  Something like `Flush` but that blocks until the reporting buffer is drained.

The `FlushLightStepTracer` call sorts of does that, but it does not block and may leave unreported spans in memory. For instance, if I call it when a [report is already in flight, the function will return immediately][1]. This means I can't block on the background flush to finish. The `Close()` function on the `Recorder` also does not drain the buffer queue.

Maybe something like:
```go
func (tr *Tracer) DrainAndClose(ctx context.Context) error { 
    // try draining the queue until `<-ctx.Done()`
}
```

Here the `context.Context` is so that I can timeout waiting on the drain to finish.

## Use case
The use case for this is for short-lived commands to be able to try and emit their spans before `os.Exit`'ing.

[1]: https://sourcegraph.com/github.com/lightstep/lightstep-tracer-go/-/blob/recorder.go#L565:20

aybabtme:
I'm happy to have a shot at this, btw. Opening this issue to see if there's interest.

yurishkuro:
fwiw Jaeger's `tracer.Close()` method behaves exactly like that, it blocks until the memory buffer is flushed.

tedsuo:
This is a duplicate of https://github.com/lightstep/lightstep-tracer-go/issues/82, plese comment there or on the associated PR if you'd like to review.

PR #85, opened 2017-06-01
Tracers can be stopped

tedsuo:
This resolves https://github.com/lightstep/lightstep-tracer-go/issues/82

* Adds a CloseTracer function
* minimal implementation for closing thrift recorder
* does not clean up thrift code or make it less racy


recorder.go
```diff
@@ -244,6 +244,23 @@ func GetLightStepAccessToken(lsTracer ot.Tracer) (string, error) {
 	}
 }
 
+type Closer interface {
```
aybabtme: Any reason to export this interface? it's the same as `io.Closer` (which could be asserted bellow) and it's not exposed anywhere in the public API, making it look like an orphan type.

recorder.go
```diff
@@ -244,6 +244,23 @@ func GetLightStepAccessToken(lsTracer ot.Tracer) (string, error) {
...
+	Close() error
+}
+
+func CloseTracer(tracer ot.Tracer) error {
```
aybabtme: The way this function is setup is not `lightstep` specific: any `basictracer.Tracer` that has a `Recorder` that can be closed will work here. So it feels to me that this generic function is odd-looking as part of this package. 

Maybe just make the `lightstep.Tracer` type implement `io.Closer`? In another piece of code where colleagues and I implement our own `Tracer` type (which originally motivated my #56), we made a wrapper interface type that clearly communicate that our `Tracer` can be closed:

```go
// Tracer is an opentracing.Tracer that can be closed.
type Tracer interface {
	io.Closer
	opentracing.Tracer
}

func New(ctx context.Context, addr string, opts ...Option) (Tracer, error) {
    // ...
}
```

In my opinion, this makes the API simpler to use and less error prone, since it only works on types that are meant to be worked with (versus doing runtime type assertions like here).

recorder.go
tedsuo: excellent point!

recorder.go
tedsuo: What is being passed in is an `ot.Tracer` because that is what the user will almost always have on hand. They could typecast to a lightstep tracer (or Closer, etc), but we do that internally in `CloseTracer` as a convenience.

thrift_rpc/recorder.go
```diff
@@ -385,6 +387,19 @@ func (r *Recorder) Flush() {
...
+	r.lock.Lock()
+	closech := r.closech
+	r.closech = nil
+	r.lock.Unlock()
+
+	if closech != nil {
+		close(closech)
+	}
+
+	return nil
```
aybabtme: This returns immediately, which means that a caller of `Close() error` does not have any idea whether the reporting loop has actually shipped its last report. This makes the `Close()` method un-useful (and a bit misleading) when someone tries to flush all transient state before, e.g, exiting a process (see #56).

Now if we consider that `Close` should block for transient reports to be shipped, we hit another problem: what if `Close` blocks for too long? Elsewhere I solved this problem by passing a `timeout` value as a configuration in the constructor (`rec.buffer` holds the pending reports):

```go
close(rec.buffer)
done := make(chan struct{})
go func() {
    wg.Wait()
    close(done)
}()
if tr.flushTimeout != 0 {
    select {
    case <-time.After(tr.flushTimeout):
        return errTimeoutFlush
    case <-done:
    }
} else {
    <-done
}
return nil
```

This gives a chance to the `Tracer` to flush its intermittent spans, while giving some control to the application over its odds of having delivered its spans. 

This may sound a bit ridiculous, but this makes all the difference when integrating tracing in CLI applications, which are very short lived. You may wonder why someone would want to trace CLI applications: turns out it's super useful for internal tools, a bit similarly to how it's super useful to single page JS applications.

thrift_rpc/recorder.go
```diff
@@ -451,19 +466,25 @@ func (r *Recorder) reportLoop() {
...
-		r.lock.Unlock()
 
-		if r.shouldFlush() {
-			r.Flush()
+			if r.shouldFlush() {
+				r.Flush()
+			}
+
+		case <-closech:
+			return
```
aybabtme: I think this should _try_ to flush a last, final report before returning.

thrift_rpc/recorder.go
tedsuo: I agree this is a race, but as I mentioned in the commit message, this PR does not attempt to make any improvements to the thrift tracer. The thrift tracer is currently scheduled for deletion. If, for some reason, the thrift tracer must have a renewed lease on life, we will improve it in a separate PR, as there are several issues with it. This code simply attempts to not break the the thrift tracer in the process of adding this feature.

thrift_rpc/recorder.go
tedsuo: See above comment regarding thrift. The thrift tracer is legacy and we do not recommend it's usage in go at this time, any new code should use the GRPC implementation. We will be adding documentation about this shortly.

tedsuo: Just to be clear: if there is a need for the thrift tracer to have a new lease on life, I will create a new PR focused on it. We do not expect users to be writing new code with it and will be EOL the thrift option shortly. If you are planing on writing new thrift based code, please let me know! (And also, please let me know why the gRPC option would not work for you.)

recorder.go
tedsuo: BTW this is only intended as a convenience function for the lightstep tracer. `Closer`, `basictracer`, etc are all incidental internal details. In the case of your custom Tracer interface, you could defer to using this to stop the lightstep tracer inside of it's Close method. We'll be removing the basictracer/recorder wrinkles soon (https://github.com/lightstep/lightstep-tracer-go/issues/86), so in the future you may be able to apply your TraceCloser interface directly to the lightstep tracer without a wrapper.

tedsuo:
Per review, we explicitly flush the the tracer on close (both grpc and thrift), and use io.Closer rather than expose a Closer interface, which was unnecessary. 

`Options.ReportTimeout` controls how long flushing can take. Worst case scenario for Close that it takes 2*ReportTimeout to finish, which would occur if the tracer had just begun to flush when Close was called (flushing was timing out).

aybabtme:
@tedsuo ah I see, didn't pick up this was Thrift only and that this recorder was meant to disappear shortly! 👍

tedsuo:
@aybabtme I added a final flush call to both anyways. :)

tedsuo:
@viruslobster I am merging this so that is does not get out of date, but I would like us to have a test harness around starting/stopping tracers before anyone uses it.

thrift_rpc/recorder.go
RaduBerinde: I was looking at using this new API and I second @aybabtme's observation. I don't understand your reply as this isn't specific to thrift. The main usecases of a cleanup API is to make sure the last batch of traces make it before exiting (e.g. `defer CloseTracer(t)` in `main`. Without blocking until the final flush completes, this API is not useful for that.

RaduBerinde: Perhaps the right thing would be to first call `FlushLightStepTracer` before closing, but that API has similar problems (#89).

RaduBerinde: To clarify - I realize the original comment was in the thrift code but I believe it applies to the gRPC path as well. I don't see any place where we are waiting for the final flush to complete. We do a final flush asynchornously (in reportLoop) but we are not waiting for it in `CloseTracer()`. In fact, we are closing the `ClientConn` immediately so the last `Flush` is likely to just exit. Am I missing something?

PR #88, opened 2017-06-09
Refactor and Ginkgo Tests

viruslobster:
This is a pretty hefty refactor, moving everything from `basictracer/*` and `thrift_rpc/*` up to the top level directory, moving shared report loop logic from the gRPC and THRIFT recorders into the Tracer implementation, and adding a Ginkgo test suite.


recorder_test.go
```diff
@@ -3,185 +3,277 @@ package lightstep
...
-	if err != nil {
-		t.Errorf("err on second close: %s", err.Error())
-	}
-}
 
-func TestDoubleCloseThrift(t *testing.T) {
-	tracer := NewTracer(Options{
-		AccessToken: "0987654321",
-		UseGRPC:     false,
+	Describe("Regarding the span buffer", func() {
```
tedsuo: No need to say "regarding" etc, just Describe("MaxBufferedSpans") is clearer.

recorder_test.go
```diff
@@ -3,185 +3,277 @@ package lightstep
...
+				It("The buffer should be empty and the capacity should be the default", func() {
+					recorder.lock.Lock()
+					defer recorder.lock.Unlock()
+					Expect(cap(recorder.buffer.rawSpans)).To(Equal(defaultMaxSpans))
+					Expect(len(recorder.buffer.rawSpans)).To(Equal(0))
+				})
+			})
+
+			Context("After spans have been started", func() {
+				It("The buffer should respond appropriately", func() {
```
tedsuo: This large It block actually covers two contexts: "When the buffer is at capacity" and "When the buffer is over capacity". Please extract these contexts, and extract appropriate `BeforeEach` blocks that set up the context.

recorder_test.go
```diff
@@ -3,185 +3,277 @@ package lightstep
...
+					}
+					recorder.lock.Lock()
+					Expect(cap(recorder.buffer.rawSpans)).To(Equal(defaultMaxSpans))
+					Expect(len(recorder.buffer.rawSpans)).To(Equal(defaultMaxSpans))
+					recorder.lock.Unlock()
+				})
+			})
+		})
+
+		Context("With a user specied buffer size", func() {
```
tedsuo: specied -> specified

recorder_test.go
```diff
@@ -3,185 +3,277 @@ package lightstep
...
 
-	_ = NewTracer(Options{
-		AccessToken: "0987654321",
-		UseThrift:   true,
-	}).(basictracer.Tracer).Options().Recorder.(*thrift_rpc.Recorder)
-}
+	Context("When calling Close() twice", func() {
+		Context("With grpc enabled", func() {
+			It("Should not fail", func() {
+				tracer = NewTracer(Options{
```
tedsuo: extract the tracer setup into a BeforeEach block.

options_test.go
```diff
@@ -1,68 +1,82 @@
 package lightstep
```
viruslobster: Ya, the exact reason I did that was I noticed the existing tests access a bunch of private fields. I can change the package, but I also would have to delete most of the existing tests

options_test.go
```diff
@@ -1,68 +1,82 @@
...
+	})
+
+	Describe("Categorizing setting trace options", func() {
+		Context("When only the TraceID is set", func() {
+			BeforeEach(func() {
+				tracer.StartSpan("x", SetTraceID(expectedTraceID)).Finish()
+			})
+
+			It("Should set the options appropriately", func() {
+				By("Only running one span")
```
viruslobster: The second by block relies on there being at least one span returned from recorder.GetSpans(). If I separate them I will have to duplicate the check made by the first by block

options_test.go
tedsuo: yeah it will be a big change, let's do that next.

options_test.go
tedsuo: ah I see that. sounds good.

recorder_test.go
viruslobster: Should a BeforeEach block be used for setup even when it only services one It block?

tedsuo: Yes, it's good practice to have setup code pulled out into a BeforeEach, especially if there is a Context block. That way, it is clearer code is being tested, vs the code that is setup and expected to work correctly.

BTW, It is okay to sometimes have small checks in your BeforeEach block, to prevent a runtime panic. For example:

```go
var span ls.Span

BeforeEach( func() {
  spans := recorder.GetSpans()
  Expect(spans).To(HaveLen(1))
  span = spans[0]
})
```

recorder_test.go
```diff
@@ -3,185 +3,285 @@ package lightstep
...
-					Value: &cpb.KeyValue_StringValue{"---------------------------------------…"},
-				},
-			},
-		}
-	}
-	return eRes
-}
+var _ = Describe("Recorder", func() {
+	const (
+		arbitraryTimestampSecs = 1473442150
```
tedsuo: In Ginkgo, it's best to declare variables and constants as close to the tests in which they are used, for clarity and to avoid accidental misuse. Move arbitraryTimestampSecs and the other top-level vars down to the tightest scope that is reasonable (most are only used in the LogRecords tests).

Also, this is just a personal style, but I prefer the single line declaration pattern for var and const, rather than the block pattern, as it often saves space.

```go
	const arbitraryTimestampSecs = 1473442150

	var recorder  *Recorder
        var tracer    ot.Tracer
	var timestamp time.Time
	var logs      []*cpb.Log
```

tedsuo:
Things are looking good @viruslobster.

I’d like to see `Recoder` dissolved into the `Tracer`. Now that there is only one implementation there’s no need for that separation. This should clean up a number of things:
- Opts and TracerConfig are combined.
- No more need for NewTracerImpl, NewTracerImplWithConfig
- Tracer should have the `Close()` method directly on it, so we can add `io.Closer` to the `Tracer` interface and simply the `CloseTracer` function.
- everything in StartSpan, startSpanWithOptions , startSpanInternal can be moved out to a `newSpan( operantionName string, tracer *tracerImpl, opts []ot.StartSpanOption)` constructor.

After that, we are going to introduce a new `opts.FinishSpanObserver` concept, which is a like a recorder that allows users to hook in their own instrumentation.

tedsuo:
This looks great @viruslobster. I made a couple of final changes:
- Cleaned up some code to make GoDocs read better.
- Removed dead code paths.
- Made the tests faster.

We should also improve documentation, but I'm fine with that being a separate PR.

bg451:
@viruslobster could you post benchmark diffs when you get the chance? To capture benchmark stats, you can do something like `go test -bench . -run none -count 5` and pipe the results into a file. I'd recommend using https://godoc.org/golang.org/x/perf/cmd/benchstat to create meaningful benchmark diffs. It's an easy way to catch any performance regression due to something like an extra allocation.

options.go
```diff
@@ -158,6 +164,9 @@ func (opts *Options) Initialize() error {
 	if opts.ReportingPeriod == 0 {
 		opts.ReportingPeriod = DefaultMaxReportingPeriod
 	}
+	if opts.MinReportingPeriod == 0 {
+		opts.ReportingPeriod = DefaultMinReportingPeriod
```
viruslobster: Should be `opts.MinReportingPeriod = DefaultMinReportingPeriod`

viruslobster:
@bg451 Here is the benchmark diff.
`name                                      old time/op  new time/op   delta
Span_Empty-8                               271ns ± 0%    368ns ± 2%    +35.59%  (p=0.008 n=5+5)
Span_100Events-8                          16.6µs ± 0%   20.8µs ± 8%    +25.60%  (p=0.008 n=5+5)
Span_1000Events-8                         17.7µs ± 5%   22.0µs ± 6%    +24.53%  (p=0.008 n=5+5)
Span_100Tags-8                            18.0µs ± 2%    8.0µs ± 4%    -55.21%  (p=0.008 n=5+5)
Span_1000Tags-8                           18.0µs ± 4%    8.1µs ± 3%    -54.98%  (p=0.008 n=5+5)
Span_100BaggageItems-8                     459µs ± 4%     43µs ± 4%    -90.55%  (p=0.008 n=5+5)
Span_100Events_100Tags_100BaggageItems-8   499µs ± 3%     64µs ± 3%    -87.25%  (p=0.008 n=5+5)
Inject_TextMap_Empty-8                    1.04µs ± 7%   0.79µs ± 3%    -23.85%  (p=0.008 n=5+5)
Inject_TextMap_100BaggageItems-8          40.8µs ± 3%    1.1µs ± 3%    -97.28%  (p=0.008 n=5+5)
Join_TextMap_Empty-8                      1.16µs ± 5%   1.21µs ± 3%     +4.66%  (p=0.032 n=5+5)
Join_TextMap_100BaggageItems-8            81.5µs ± 4%    1.8µs ± 3%    -97.82%  (p=0.008 n=5+5)
StringsRepeatBaseline-8                   12.1µs ± 2%   33.8µs ±20%   +179.98%  (p=0.008 n=5+5)
ShortMessage-8                            19.5ns ± 7%  639.6ns ±16%  +3173.29%  (p=0.008 n=5+5)
LongMessage-8                             12.8µs ± 7%   26.9µs ±17%   +110.54%  (p=0.008 n=5+5)`

span.go
```diff
@@ -31,19 +18,80 @@ type spanImpl struct {
...
+
+	// Look for a parent in the list of References.
+	//
+	// TODO: would be nice if we did something with all References, not just
+	//       the first one.
+ReferencesLoop:
+	for _, ref := range opts.Options.References {
+		switch ref.Type {
+		case ot.ChildOfRef,
+			ot.FollowsFromRef:
```
jmacd: nit: put FollowsFrom and ChildOfRef on one line

options.go
```diff
@@ -1,43 +1,270 @@
...
+		opts.MinReportingPeriod = DefaultMinReportingPeriod
+	}
+	if opts.ReportTimeout == 0 {
+		opts.ReportTimeout = DefaultReportTimeout
+	}
+	if opts.ReconnectPeriod == 0 {
+		opts.ReconnectPeriod = DefaultReconnectPeriod
+	}
+	if opts.Tags == nil {
+		opts.Tags = make(map[string]interface{})
```
jmacd: map[string]interface{}{}

tracer.go
```diff
@@ -0,0 +1,357 @@
...
+
+// NewTracer creates and starts a new Lightstep Tracer.
+func NewTracer(opts Options) Tracer {
+	err := opts.Initialize()
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil
+	}
+
+	attributes := make(map[string]string)
```
jmacd: ditto...

grpc_tracer_test.go
```diff
@@ -0,0 +1,456 @@
...
+	. "github.com/lightstep/lightstep-tracer-go"
+	cpb "github.com/lightstep/lightstep-tracer-go/collectorpb"
+	cpbfakes "github.com/lightstep/lightstep-tracer-go/collectorpb/collectorpbfakes"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	ot "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+)
+
+var _ = Describe("Tracer", func() {
```
jmacd: This test is :100:

tedsuo:
Looks great. Thanks @viruslobster for all the hard work!

Issue #89, opened 2017-06-21
FlushLightStepTracer is not reliable


RaduBerinde:
We have a situation where we are tracing some short-lived tests and some traces aren't making it to lightstep. We added a call to `FlushLightStepTracer` and that seems to help, but it still doesn't seem reliable. Sometimes we see this error:
```
I170621 15:52:43.061339 50 util.go:39  LightStep instrumentation error (a previous Report is still in flight; aborting Flush()). Set the Verbose option to enable more logging.
```

The flush code indeed just bails if a `Report` is in progress. This is problematic on two fronts:
 - we won't flush any traces that were generated since we started this last Report
 - more importantly - normally, the `FlushLightStepTracer` call returns only after we've flushed, so it's reasonable to call this before exiting. But if we hit this error, we won't even wait for the in-progress Report to finish.

I think that this case should be handled by the flush function - it should wait until the current Report finishes and then start another one.

RaduBerinde:
A possible partial "fix" would be to at least return the error from `Flush()` so the caller can do a retry loop.

tedsuo:
Hi @RaduBerinde I see the case you are discussing - flush should wait for the current report to complete, and possibly send another report if there is one in flight.

We can also add an error in the case where the collector connection failing. However, it is unclear to me that we should encourage retries on error - any retry logic that has a hope of being fruitful should be internal to the tracer. Also, FYI we are adding a centralized error handler for reporting purposes: https://github.com/lightstep/lightstep-tracer-go/issues/91

RaduBerinde:
I agree, it would be preferable to not rely on the caller retrying.

RaduBerinde:
By the way, I think `CloseTracer` has a similar problem - I posted about it in this PR (unfortunately, a while after it was merged): https://github.com/lightstep/lightstep-tracer-go/pull/85#discussion_r123800322

tedsuo:
Sorry, I missed your comments on the previous PR! But I agree with you - Close has the same problem. Also, we have internal errors with no reporting mechanism. So I think we can solve these issues together in v0.14:

- Centralize error handling with `OnError` handler
- `Flush` ensures messages in buffer prior to invocation are reported.
- `Close` ensures buffer is empty before returning.
- `Close` and `Flush` can be aborted via `context.Context`

```go
import ot "github.com/opentracing/opentracing-go"

// Backwards compatible convenience methods. The only error returned 
[...]

PR #97, opened 2017-07-19
Reliable Flushing

viruslobster:
FlushLightstepTracer should retry flushing if there is a previous report in flight instead of just bailing.

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -77,6 +77,7 @@ type tracerImpl struct {
 	flushing reportBuffer
 
 	// Flush state.
+	flushingLock      sync.Mutex
 	reportInFlight    bool
 	lastReportAttempt time.Time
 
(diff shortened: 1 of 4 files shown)


viruslobster:
This PR is meant to address #89

tracer.go
```diff
@@ -253,9 +253,16 @@ func (r *tracerImpl) Flush() {
 	}
 
 	if r.reportInFlight == true {
-		maybeLogError(errPreviousReportInFlight, r.opts.Verbose)
 		r.lock.Unlock()
-		return
+		tickerChan := time.Tick(r.opts.MinReportingPeriod / 2)
```
viruslobster: Any thoughts on the period here? Also, should we add a max retry and then fail with an error?

tracer_test.go
```diff
@@ -1,20 +1,79 @@
...
 
+	. "github.com/lightstep/lightstep-tracer-go"
 	cpb "github.com/lightstep/lightstep-tracer-go/collectorpb"
 	cpbfakes "github.com/lightstep/lightstep-tracer-go/collectorpb/collectorpbfakes"
 	"github.com/lightstep/lightstep-tracer-go/lightstepfakes"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+// type Connection interface {
```
viruslobster: woops, gonna remove this

tracer.go
tedsuo: as per our discussion, maybe channels will help:
```go
reportCh := r.reportInFlight
r.lock.Unlock()

if reportCh != nil {
  <-reportCh
}

r.lock.Lock()
r.reportInFlight = true
r.lock.Unlock()

resp, err := r.client.Report(ctx, &r.flushing)

r.lock.Lock()
reportCh := r.reportInFlight
r.reportInFlight = nil
r.lock.Unlock()

close(reportCh)
```

tracer.go
```diff
@@ -239,27 +240,20 @@ func (r *tracerImpl) RecordSpan(raw RawSpan) {
 
 // Flush sends all buffered data to the collector.
 func (r *tracerImpl) Flush() {
+	r.flushingLock.Lock()
+	defer r.flushingLock.Unlock()
 	r.lock.Lock()
+	defer r.lock.Unlock()
```
tedsuo: This is dangerous. `defer r.lock.Unlock()` implies to other programmers that the entire function is inside the `r.lock`, which is not the case, and causes an extra call to `Unlock()` in one case, which is very dangerous. (you unlock once, other code locks, you unlock again, race occurs). 

I would suggest extracting two functions, `beforeFlush() (*reportBuffer, bool) ` and `afterFlush(error)`, which encapsulate the two sets of buffer/memory management. That way, in each function you can use `defer r.lock.Unlock()` and it's cleaner.

viruslobster:
don't know how I missed that... Fixed and added a test

PR #100, opened 2017-08-10
Buffer recovery on flushing error

viruslobster:
Fixed the call to copy in reportBuffer::mergeFrom

### DIFF
--- a/report_buffer.go
+++ b/report_buffer.go
@@ -1,6 +1,8 @@
 package lightstep
 
-import "time"
+import (
+	"time"
+)
 
 type reportBuffer struct {
 	rawSpans             []RawSpan
@@ -69,7 +71,10 @@ func (into *reportBuffer) mergeFrom(from *reportBuffer) {
 		space = unreported
 	}
 
+	// must have the len to accommodate new spans for copy to work
+	into.rawSpans = append(into.rawSpans, make([]RawSpan, space)...)
 	copy(into.rawSpans[have:], from.rawSpans[0:space])
+
 	into.droppedSpanCount += int64(unreported - space)
 
 	from.clear()
--- a/tracer_test.go
+++ b/tracer_test.go
@@ -1,6 +1,7 @@
 package lightstep_test
 
 import (
+	"errors"
 	"sync"
 	"time"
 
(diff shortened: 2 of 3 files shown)


report_buffer.go
```diff
@@ -69,7 +71,10 @@ func (into *reportBuffer) mergeFrom(from *reportBuffer) {
 		space = unreported
 	}
 
+	// must have the len to accommodate new spans for copy to work
+	into.rawSpans = append(into.rawSpans, make([]RawSpan, space)...)
 	copy(into.rawSpans[have:], from.rawSpans[0:space])
+
```
tedsuo: I think this may work, avoiding the copy step:
```go
into.rawSpans = append(into.rawSpans, from.rawSpans[0:space]...)
```

version.go
```diff
@@ -1,2 +1,3 @@
 package lightstep
+
 const TracerVersionValue = "0.13.0"
```
tedsuo: this file is auto-generated, best not to touch it.

tracer_test.go
```diff
@@ -28,16 +29,16 @@ var _ = Describe("SpanRecorder", func() {
 			cancelch = make(chan struct{})
 			startTestch = make(chan bool)
 			fakeClient = new(cpbfakes.FakeCollectorServiceClient)
-			tracer = NewTracer(Options{
-				AccessToken: "YOU SHALL NOT PASS",
-				ConnFactory: fakeGrpcConnection(fakeClient),
-			})
```
tedsuo: Why was the tracer pushed down, btw?

tracer_test.go
```diff
@@ -68,15 +73,49 @@ var _ = Describe("SpanRecorder", func() {
...
+			BeforeEach(func() {
+				returnErr := true
+
+				fakeClient.ReportStub = func(ctx context.Context, req *cpb.ReportRequest, options ...grpc.CallOption) (*cpb.ReportResponse, error) {
+					if returnErr {
+						returnErr = false
+						return nil, errors.New("fail")
+					}
+					return new(cpb.ReportResponse), nil
+				}
```
tedsuo: A cleaner way to do this with counterfieter is to override specific calls:

```go
fakeClient.ReportReturnsOnCall(0, nil, errors.New("fail"))
fakeClient.ReportReturns(new(cpb.ReportResponse), nil)
```

tracer_test.go
viruslobster: because for the new test I had to specify MinReportingPeriod

tedsuo:
LGTM!

PR #101, opened 2017-08-11
Centralized Error Handling 

viruslobster:


tracer.go
```diff
@@ -181,153 +205,167 @@ func (t *tracerImpl) Extract(format interface{}, carrier interface{}) (ot.SpanCo
 	return nil, ot.ErrUnsupportedFormat
 }
 
-func (r *tracerImpl) reconnectClient(now time.Time) {
-	conn, err := r.client.ConnectClient()
+func (t *tracerImpl) reconnectClient(now time.Time) {
+	conn, err := t.client.ConnectClient()
 	if err != nil {
-		maybeLogInfof("could not reconnect client", r.opts.Verbose)
```
tedsuo: we should leave the logging in until we have another solution.

tracer.go
```diff
@@ -181,153 +205,167 @@ func (t *tracerImpl) Extract(format interface{}, carrier interface{}) (ot.SpanCo
...
-	r.flushingLock.Lock()
-	defer r.flushingLock.Unlock()
+func (t *tracerImpl) Flush(ctx context.Context) {
+	t.flushingLock.Lock()
+	defer t.flushingLock.Unlock()
 
-	err := r.preFlush()
+	err := t.preFlush()
 	if err != nil {
-		maybeLogError(err, r.opts.Verbose)
```
tedsuo: leave logging in for now

tedsuo:
Also, the more I think about it, I would prefer not to change the initialization interface to return an error at this point, as that will make this into a breaking change. Instead, please add `Validate() error` method to the `Options` object. In a future commit, we can break the constructor.

tedsuo:
For Context, the logic we are looking for is "if a Context is present, let it drive." This means we will continue to retry indefinitely until the Context is Done.

viruslobster:
adding a `Validate() error` method to opts gets around users not knowing of there options are valid, but even if they are valid, NewTracer can still return nil if we fail to establish a connection:
```
conn, err := impl.client.ConnectClient()
	if err != nil {
		impl.onError(err)
		return nil
	}
```

viruslobster:
Also, what exactly do you mean by "if a context is present"? Do you want users to be able to pass in nil for a context?

PR #105, opened 2017-09-18
Delete local *.pb.go files and reference the proto repo.

MatthewDolan:


MatthewDolan:
@danielamiao good call! Bumped the version number and added an entry change log.

propagation_ls.go
```diff
@@ -4,7 +4,7 @@ import (
 	"encoding/base64"
 
 	"github.com/golang/protobuf/proto"
-	lightstep "github.com/lightstep/lightstep-tracer-go/lightsteppb"
+	lightstep "github.com/lightstep/api-public/go/lightstep"
```
danielamiao: Does this need to be updated to `lightstep-tracer-protos`?

MatthewDolan: Oops. Good catch!

collectorpb/collector.go
```diff
@@ -1,10 +1,18 @@
 package collectorpb
 
-func (res *ReportResponse) Disable() bool {
-	for _, command := range res.GetCommands() {
+import (
+    "github.com/lightstep/lightstep-tracer-protos/go/lightstep/collector"
+)
+
+type ReportResponseWrapper struct {
+    *collectorpb.ReportResponse
```
tedsuo: This wrapper is unnecessary. Can we move this method over to `lightstep-tracer-protos` as well, rather than have this spread out between two repos and then wrapped? `Disable()` is part of the RepostResponse API.

MatthewDolan: I'm confused what you mean when you say "Disable() is part of the RepostResponse API" It isn't a part of the generated code that protoc create, right? I was trying to avoid augmenting a package with generated code in it.

MatthewDolan: Discussed offline, let's follow up in another PR if necessary.

PR #110, opened 2017-09-25
Small clean up and typo fixes.

MatthewDolan:


interfaces.go
```diff
@@ -12,8 +12,8 @@ type Connection interface {
 	io.Closer
 }
 
-// ConnectorFactory is for testing purposes.
-type ConnectorFactory func() (interface{}, Connection, error)
+// ConnectionFactory is for testing purposes.
+type ConnectionFactory func() (interface{}, Connection, error)
```
MatthewDolan: This is a backward incompatible change, but since it's only for testing, it should probably be fine. It might make sense to just revert it instead of risking breaking builds.

danielamiao: It's your call, I am generally OK with backwards incompatible changes as long as we version bump.

PR #112, opened 2017-09-26
Add http collector client

MatthewDolan:


options.go
```diff
@@ -130,11 +130,12 @@ type Options struct {
 	// Set Verbose to true to enable more text logging.
 	Verbose bool `yaml:"verbose"`
 
-	// DEPRECATED: set `UseThrift` to true if you do not want gRPC
-	UseGRPC bool `yaml:"usegrpc"`
-
-	// Switch to
+	// Force the use of a specific transport protocol.
+	// If multiple are set to true, the top one is chosen.
+	// If none are set to true, GRPC is defaulted to.
```
danielamiao: Let's be explicit about ordering thrift -> http -> grpc in the comments here since one might accidentally re-order them in the code below.

options.go
```diff
@@ -130,11 +130,12 @@ type Options struct {
...
 	UseThrift bool `yaml:"use_thrift"`
+	UseHttp bool `yaml:"use_http"`
+	UseGRPC bool `yaml:"usegrpc"`
```
danielamiao: NIT: is there a reason this isn't `use_grpc` for consistency?

proto_converter.go
```diff
@@ -0,0 +1,186 @@
...
+	"time"
+)
+
+type protoConverter struct {
+	verbose        bool
+	maxLogKeyLen   int // see GrpcOptions.MaxLogKeyLen
+	maxLogValueLen int // see GrpcOptions.MaxLogValueLen
+}
+
+func newProtoConverter(options Options) *protoConverter {
```
danielamiao: This entire file makes me so happy 👍  to _more_ organization!

http_collector_client.go
```diff
@@ -0,0 +1,140 @@
+package lightstep
```
danielamiao: This entire file's format seems a bit off to me on github, maybe run `go fmt` again?

options.go
```diff
@@ -130,11 +130,12 @@ type Options struct {
...
 
-	// DEPRECATED: set `UseThrift` to true if you do not want gRPC
-	UseGRPC bool `yaml:"usegrpc"`
-
-	// Switch to
+	// Force the use of a specific transport protocol.
+	// If multiple are set to true, the top one is chosen.
+	// If none are set to true, GRPC is defaulted to.
 	UseThrift bool `yaml:"use_thrift"`
+	UseHttp bool `yaml:"use_http"`
```
danielamiao: Do we plan to support HTTP/JSON in the go tracer in the future? I imagine there will be an additional option for use_http_json vs use_http_proto?

MatthewDolan: I doubt it. Go has good support for proto and I don't this it's a big deal to rely on proto as a dependency. http/json is intended for languages that don't have good proto support (like for example javascript has in the past had bad proto support because it's very slow at bit packing operations).

options.go
MatthewDolan: I am just copying an already existing field. Changing the yaml would be backward incompatible and mean I would need to do a major instead of a minor release. I would like to avoid that.

options.go
MatthewDolan: +1

http_collector_client.go
MatthewDolan: done!

proto_converter.go
MatthewDolan: 😄

PR #119, opened 2017-10-17
Events

tedsuo:
To fully integrate the lightstep tracer with application diagnostic tools such as metrics and logging. 

## CHANGELOG
- Tracer no longer logs to stdout by default.
- Verbose option is now deprecated.
- Added SetGlobalEventHandler and Event types.
- Added EventStatusReport to provide client metrics.
- Close and Flush no longer return errors.
- Close and Flush now take a context parameter.
- Options can be validated.
- SpanRecorder only blocks caller, not tracer internals.
- Tracerv0_14 added for backwards compatibility

## Documentation
Please review the documentation to get a sense of this new feature, and ensure it is understandable. Please look at the GoDocs in particular.

- New README sections “Flushing and Closing” and “Event Handling”
- Visit https://github.com/lightstep/lightstep-tracer-go/tree/tedsuo/on-event#flushing-and-closing-managing-the-tracer-lifecycle to view the README.
- Godoc documentation and layout should be easy to understand. 
- Open this branch locally and run `godoc -http :8080`
- Visit http://localhost:8080/pkg/github.com/lightstep/lightstep-tracer-go to view the godocs.

# Event Handling

Events are emitted by the LightStep tracer as a reporting mechanism. They are handled by adding an `OnEvent` callback to the `Options` passed to `NewTracer`. Events may be cast to specific event types in order access additional information.

`OnEvent` sets a global handler to receive tracer events as they occur. The Event handler will receive events from any active tracers, as well as errors in static functions.  It is suggested that you set your OnEvent handler before starting your tracer, but it is safe to set a new handler at any time. Setting a new handler removes the previous one.

The `OnEvent` handler is called synchronously – do not block in your handler as it may interfere with the tracer. If no `OnEvent` handler is provided, a `LogOnceOnError` handler is used by default. `NewOnEventLogger`, `NewOnEventLogOneError`, and `NewOnEventChannel` are standard handlers available as a convenience.

NOTE: Event handling is for reporting purposes only. It is not intended as a mechanism for controlling or restarting the tracer. Connection issues, retry logic, and other transient errors are handled internally by the tracer.  

NOTE: To ensure that events can be accurately identified, each event type contains a sentinel method matching the name of the type. This method is a no-op, it is only used for type coersion.

## Why a global OnEvent handler? Why not return errors from method calls?

In Go (as in other programming languages) synchronous methods that return their results are good, while global- or package- level state is frowned upon. However, in this case, a global event handler solves several problems:

- Most tracing events have no mechanism for being returned directly to the caller.
- Events may occur inside of package-level functions, or deep within tracer subsystems. A global handler allows these constructs to emit events without the need to thread a reporter into hard to get areas.
- Setup for metrics and logging may occur separately from tracer setup. Decoupling event handling from tracer setup makes this easier.
- Returned errors on Close and Flush were encouraging users to handle the errors by writing their own retry and backoff logic. We want users to report errors, but not handle them in this manner.
- Having a mix of returned errors for tracer methods, logging for static functions, and event handling for tracer internals creates a messy surface area. Better to return no errors anywhere, and pass them all to the event handler.

## UNRELATED FIX: SpanRecorder only blocks caller, not tracer internals

I noticed that the `SpanRecoder` was being called inside of the lock during `RecordSpan`. This means that any blocking which occurs in the 3rd-party recorder will block the entire tracer, rather than just that particular caller. I couldn’t imagine a scenario where that would be preferred, and it was a small change, so I fixed it.


README.md
```diff
@@ -12,30 +12,93 @@ The LightStep distributed tracing library for Go.
 $ go get 'github.com/lightstep/lightstep-tracer-go'
 ```
 
-## Getting started
```
MatthewDolan: nit: It would be nice to split README change into a separate PR. But I don't have strong feelings about that.

event_handlers.go
```diff
@@ -0,0 +1,100 @@
...
+// the previous one.
+//
+// OnEvent is called synchronously – do not block in your handler as it may interfere
+// with the tracer. If no OnEvent handler is provided, a LogOnceOnError handler
+// is used by default.
+//
+// NOTE: Event handling is for reporting purposes only. It is not intended as a
+// mechanism for controling or restarting the tracer. Connection issues, retry
+// logic, and other transient errors are handled internally by the tracer.
+func OnEvent(handler OnEventHandler) {
```
MatthewDolan: I would prefer a slightly more explicit name here (e.g. `SetGlobalOnEventHandler(...)`). Similar to how we have `SetGlobalTracer(...)`

event_handlers.go
```diff
@@ -0,0 +1,100 @@
...
+	eventHandlerLock.Lock()
```
MatthewDolan: This lock isn't necessary, right? `atomic.Value` should be thread-safe.

events.go
```diff
@@ -0,0 +1,290 @@
...
+type eventStatusReport struct {
+	startTime      time.Time
+	finishTime     time.Time
+	sentSpans      int
+	droppedSpans   int
+	encodingErrors int
+}
+
+func newEventStatusReport(startTime, finishTime time.Time, sentSpans, droppedSpans, encodingErrors int) *eventStatusReport {
+	return &eventStatusReport{startTime: startTime, finishTime: finishTime, sentSpans: sentSpans, droppedSpans: droppedSpans, encodingErrors: encodingErrors}
```
MatthewDolan: nit: lines too long.

events.go
```diff
@@ -0,0 +1,290 @@
...
+func (s *eventStatusReport) DroppedSpans() int {
+	return s.droppedSpans
+}
+
+func (s *eventStatusReport) EncodingErrors() int {
+	return s.encodingErrors
+}
+
+func (s *eventStatusReport) String() string {
+	return fmt.Sprint("STATUS REPORT start: ", s.startTime, ", end: ", s.finishTime, ", dropped spans: ", s.droppedSpans, ", encoding errors: ", s.encodingErrors)
```
MatthewDolan: nit: lines too long.

options.go
```diff
@@ -222,6 +227,20 @@ func (opts *Options) Initialize() error {
 	return nil
 }
 
+// Validate checks that all required fields are set, and no options are incorrectly
+// configured.
+func (opts *Options) Validate() error {
```
MatthewDolan: nit: It might be nice to have this be a []error instead so you can see all the errors that validate returns. Don't feel strongly about this though.

proto_converter.go
```diff
@@ -114,12 +115,7 @@ func (converter *protoConverter) toField(key string, value interface{}) *cpb.Key
 		field.Value = &cpb.KeyValue_BoolValue{BoolValue: reflectedValue.Bool()}
 	default:
 		field.Value = &cpb.KeyValue_StringValue{StringValue: fmt.Sprint(reflectedValue)}
-		maybeLogInfof(
-			"value: %v, %T, is an unsupported type, and has been converted to string",
-			converter.verbose,
-			reflectedValue,
-			reflectedValue,
-		)
+		onEvent(newEventUnsupportedValue(key, value))
```
MatthewDolan: I like this syntax a lot more!

thrift_collector_client.go
```diff
@@ -91,7 +91,6 @@ func (client *thriftCollectorClient) ConnectClient() (Connection, error) {
 	} else {
 		transport, err := thrift.NewTHttpPostClient(client.collectorURL, client.reportTimeout)
 		if err != nil {
-			maybeLogError(err, client.verbose)
```
MatthewDolan: Replace with an `onEvent(...)`?

tracer_helpers.go
```diff
@@ -0,0 +1,59 @@
...
+	lsTracer, ok := tracer.(Tracer)
+	if !ok {
+		return "", newEventUnsupportedTracer(tracer)
+	}
+
+	return lsTracer.Options().AccessToken, nil
+}
+
+// DEPRECATED: use Flush instead.
+func FlushLightStepTracer(lsTracer opentracing.Tracer) error {
```
MatthewDolan: nit: call out to `Flush(...)` instead of copying the code.

tracer_helpers.go
```diff
@@ -0,0 +1,59 @@
...
+	if !ok {
+		return newEventUnsupportedTracer(lsTracer)
+	}
+
+	tracer.Flush(context.Background())
+	return nil
+}
+
+// DEPRECATED: use Close instead.
+func CloseTracer(tracer opentracing.Tracer) error {
```
MatthewDolan: Same comment as above. Call out to `Close(...)`

event_handlers.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+var eventHandler atomic.Value
+var eventHandlerLock sync.Mutex
+
+// An OnEventHandler can be registered with OnEvent to
+type OnEventHandler func(Event)
+
+// onEvent is a thread-safe function for emiting tracer events.
+func onEvent(event Event) {
```
MatthewDolan: nit: It might be nice to add

```
if event == nil {
  return
}
```

So that callers don't necessary have to check for nil.

README.md
```diff
@@ -12,30 +12,93 @@ The LightStep distributed tracing library for Go.
...
+  case ErrorEvent:
+    logger.Error("LS Tracer error: %s", event)
+  default:
+    logger.Info("LS Tracer info: %s", event)
+  }
+}
+
+func main() {
+  // setup event handler first to catch startup errors
+  lightstep.OnEvent(logAndMetricsHandler)
```
jmacd: Why not specify the handlers on the NewTracer call?

event_handlers.go
jmacd: The fact that this can only be called once makes me think it should be provided to the NewTracer constructor. Otherwise, any rogue library could install its own handler. This is something I believe should be set in main, why make it possible to change?

event_handlers.go
```diff
@@ -0,0 +1,100 @@
...
+
+type logOneError struct {
+	sync.Once
+}
+
+func (l *logOneError) OnEvent(event Event) {
+	switch event := event.(type) {
+	case ErrorEvent:
+		l.Once.Do(func() {
+			log.Printf("LS Tracer error: (%s). NOTE: Set the Verbose option to enable more logging.\n", event.Error())
```
jmacd: The NOTE has to be removed ^^^.

options.go
```diff
@@ -127,7 +133,9 @@ type Options struct {
 	// DropSpanLogs turns log events on all Spans into no-ops.
 	DropSpanLogs bool `yaml:"drop_span_logs"`
 
-	// Set Verbose to true to enable more text logging.
+	// DEPRECATED: the LS tracer no longer logs to stdout. See the documentation
```
jmacd: Slightly incorrect. "DEPRECATED: The LightStep library prints the first error to stdout by default. To print subsequent messages, register ` lightstep.OnEvent(NewOnEventLogger())`.

events.go
```diff
@@ -0,0 +1,290 @@
...
+const (
+	FlushErrorTracerClosed   EventFlushErrorState = "flush failed, the tracer is closed."
+	FlushErrorTracerDisabled EventFlushErrorState = "flush failed, the tracer is disabled."
+	FlushErrorTransport      EventFlushErrorState = "flush failed, could not send report to Collector"
+	FlushErrorReport         EventFlushErrorState = "flush failed, report contained errors"
+)
+
+var (
+	flushErrorTracerClosed   = errors.New(string(FlushErrorTracerClosed))
+	flushErrorTracerDisabled = errors.New(string(FlushErrorTracerDisabled))
```
jmacd: Why not make EventFlushErrorState implement error? Just a 
```
func (s EventFlushErrorState) Error() string { return string(s) }
```
and you can use them as errors, remove these two vars.

proto_logencoder.go
```diff
@@ -76,6 +76,7 @@ func (lfe *grpcLogFieldEncoder) EmitObject(key string, value interface{}) {
 	lfe.emitSafeKey(key)
 	jsonBytes, err := json.Marshal(value)
 	if err != nil {
+		onEvent(newEventUnsupportedValue(key, value))
```
jmacd: The user is going to want the error, I think.

proto_converter.go
jmacd: This offends me. 😃 

Could we change it? (I know this is the old behavior.)
If the call to Sprint panics, then we might call it unsupported. Otherwise, it may implement fmt.Stringer in which case you're putting a spurious warning onto the user. If it prints but is not a Stringer, I will still argue it's supported based on the philosophy that it should not be an error to instrument things.

jmacd:
About the unrelated fix: there was never an intention to support "third-party" recorders. That option was left-over from the basictracer package. Holding a lock seems OK when you control both sides of the code, but if it's not needed it's not needed.

README.md
tedsuo: I'm happy to split all of this out if it's hard to review in one go.

event_handlers.go
MatthewDolan: Or updated!

event_handlers.go
tedsuo: This needs more investigation in general, I'm still looking into the pitfalls of `atomic.Value`.

options.go
tedsuo: I would make a `type ValidationError []error` for this. I can copy that code over: https://godoc.org/code.cloudfoundry.org/bbs/models#ValidationError

I thought about adding it, but didn't because KISS and we could add it later.

README.md
tedsuo: We emit events from static methods and other places that cannot be reached easily, except by a global handler. Also, reporting validation errors via an invalid `Options` object felt wrong.

event_handlers.go
tedsuo: @jmacd please see the part of the PR description called "Why a global OnEvent handler?" for my reasoning.

jmacd: There are two questions there, one "Why" and one "Why not". There's only one answer for the Why I can see:

_Events may occur inside of package-level functions, or deep within tracer subsystems. A global handler allows these constructs to emit events without the need to thread a reporter into hard to get areas._

This to me is suspect. We already have a global tracer, and if that global tracer is a lightstep tracer, then we can instead get the global tracer, see if it's a lightstep tracer, then cast it, get the event handler, and use it.

Having the event handler be registered by a global function just seems to invite misuse. Registering it when the tracer starts feels less prone to surprise.

tedsuo: I would be down with `SetEventHandler` or something similar.

event_handlers.go
tedsuo: Good catch

event_handlers.go
MatthewDolan: I tend to agree. It seems like having a global is asking for action at a distance and confusing outcomes.

events.go
tedsuo: So, the "implementation of error" happens one level higher, in `EventFlushError`. `EventFlushError` contains two things: the category of error (which are enumerated by `EventFlushErrorState`) and the underlying error. Some error states (`FlushErrorTransport` and  `FlushErrorReport`) could contain a variety of errors that are generated elsewhere: 

```go
  event := newEventFlushError(someNetworkError, FlushErrorTransport)
```

The tracer itself generates the "Flush on Close" and "Flush on Disabled" errors, so those errors happen to live in this package:

```go
  event := newEventFlushError(flushErrorTracerClosed, FlushErrorTracerClosed)
```

The idea was to have flush errors be identifiable on three levels: As an ErrorEvent, a FlushErrorEvent, and type of flush error (EventFlushErrorState). Having each flush error by a separate event type (`EventFlushErrorTracerClosed`, `EventFlushErrorTracerDisabled`, etc) seemed too fine grained... you can't easily identify something as a `FlushError` which is an important distinction. But those type semantics are what I'd like vetted the most... is this sliced up in the most useful way for users to alert and report effectively?

options.go
tedsuo: Good catch

proto_converter.go
tedsuo: I forgot to mention this! Rather than always throw this, I was wondering if we should explicitly typecast to `Stringer`, and record that value without a fuss. I would assume that recording the `Stringer` value was the intention of the caller. Likewise, we should check for the `error` interface, as that has a common and intentional string representation. We should only emit this event when we get something without any known common representation - and then record the `fmt.Sprint` value for it. 

Bear in mind, we are not "throwing an exception" or otherwise exploding in the users face, we are alerting the human in a safe way that they are doing something they probably did not intend to do. They can ignore these warnings if they don't care. If your argument is that it simply isn't an error to record an object, even an "unrecordable" one, as that was intentional on the part of the user – they want to introspect the damn thing, for example – then yes, I agree the warning should go away.  But bear in mind, the OT docs say: 

```
    // Tag values can be numeric types, strings, or bools. The behavior of
    // other tag value types is undefined at the OpenTracing level. If a
    // tracing system does not know how to handle a particular value type, it
    // may ignore the tag, but shall not panic.
```

So this event is to inform users that they are "out of compliance" with OT. My preference is that users should choose how they want to introspect an object when they record it:

```go
// This is an intentional representation; output is stable regardless of tracing vendor.
tracer.Tag("someObjTag", fmt.Sprintf("%T", someObj))

// In OT, you have no clue what data this will output – or even if the tag will be added.
tracer.Tag("someObjTag", someObj)
```

If we don't want to have this bikeshed, I can just eliminate this event for now, and add it in a subsequent PR.

proto_logencoder.go
tedsuo: Good point.

thrift_collector_client.go
tedsuo: For better or worse, `ConnectClient` errors are currently getting emitted on the caller's side.

tracer_helpers.go
tedsuo: Can't if we want to preserve original behavior (returning a typecheck error).

event_handlers.go
tedsuo: Totally understand the critique! I too, hate globals and find them suspicious. It is possible to move event handling back to If we want to allow event reporting be an `Option` on a Tracer instead of a global handler, there are three things we have to do:

- cease to emit events for typecast errors in `lightstep.Flush` and `lightstep.Close`.
- refactor `collectorClient` implementations to take an event handler.
- live with the oddness around using an invalid Options object to report startup failures.

The second two are doable. Ceasing to emit events though... I'm concerned about that. That gets at the core design requirements of this feature, which is to report diagnostics while no longer returning errors from lightstep functions.

An example of this situation btw, just so it's clear that it's worth thinking about: a customer instruments all of their code with LightStep. But they they read a cool OpenTracing article and wrap the lightstep tracer up in some kind of meta tracer. Then their flushing and shutdown code ceases to function. That means dropped spans. When we silently dropped spans on shutdown in the past, users were pretty mad when they eventually noticed.

```go
func startTracer(){
  lsTracer := lightstep.NewTracer(opts)

  // when they wrap our tracer up
  logger := glog.NewLogger()
  tracer := ot_helpers.NewLogTeeTracer(lsTracer, logger)

  // and register the wrapped tracer...
  opentracing.SetGlobalTracer(tracer)
}

func stopTracer(ctx context.Context){
  // ...shutdown will start failing here. Is this a silent failure? How do we notify?
  lightstep.Close(ctx, opentracing.GlobalTracer())
}
```

A relevant design input is that users have reacted poorly to there being `error` return values on lightstep functions. When users see an error, they want to "handle" it somehow... it's been difficult to explain to users that they should alert on errors but otherwise ignore them. It's been enough of a problem that one of the goals was to unify the reporting mechanism in such a way that it would not encourage users to try to create a feedback loop with the tracer. For example, receiving a flush error and trying to manually close and restart the tracer, rather than letting the tracer handle it. Patterns like that have lead to users writing code that blocks shutdown, and other problems.

The alternative solution of grabbing at the global tracer to get an event handler won't work in the situation above. I would suggest we go with an explicit global event handler, or a solution for the above issues which does not require any globals.

Hope that helps!

options.go
MatthewDolan: That's reasonable. It feels like a great to have feature but no where close to a requirement.

thrift_collector_client.go
MatthewDolan: 👍

event_handlers.go
tedsuo: Looks like no lock is needed here. Fun fact: using a typed function type rather than an interface type avoids a panic in atomic.Value.

```go
// this works well with atomic.Value
type EventHandler func(Event)

// using an interface instead causes atomic.Value to panic when you
// store a new value with a different underlying concrete type.
type EventHandler interface {
  OnEvent(Event)
}

// sticking the EventHandler interface in a stable concrete type 
// avoids that issue. 
type eventHandlerContainer struct {
  handler EventHandler
}
```

^ That was a little weird to me... when embedded, the interface object is treated as memory layout. On it's own, it's treated as language construct and the pointer to the concrete type is walked.

proto_converter.go
jmacd: You make a good case. I'd like to not print this for fmt.Stringer, assuming that's intentional, otherwise fine and I retract my offended remarks.

events.go
jmacd: 👍

event_handlers.go
jmacd: I stand convinced. Ted's explanation revolves around corner cases that users legitimately fall into, like retrying a lightstep call that would fail and never notify the user (because a different tracer is installed) if we didn't make the event notifier global.

README.md
jmacd: :+1: The case about `lightstep.Close` being a non-OT method and how users do the wrong thing when presented with an error convinced me.

tedsuo:
I changed `OnEvent` to `SetGlobalEventHandler`, and I believe I addressed all of the nits.

interfaces.go
```diff
@@ -34,15 +34,17 @@ type SpanRecorder interface {
...
+// Tracer extends the `opentracing.Tracer` interface with methods for manual
+// flushing and closing. To access these methods, you can take the global
+// tracer and typecast it to a `lightstep.Tracer`. As a convenience, the
+// lightstep package provides static functions which perform the typecasting.
 type Tracer interface {
 	ot.Tracer
 
 	// Close flushes and then terminates the LightStep collector
-	Close() error
+	Close(context.Context)
```
MatthewDolan: Just noticed this, adding context here is a breaking API change, right? Do we want to do that? Would it be better to add new methods (`CloseContext(...)` and `FlushContext(...)`). I have noticed that pattern in the core go code when they want to add context but don't want it to be backward incompatible.

tracer_helpers.go
MatthewDolan: Ah I see.

One thought, adding an `error` to the return of `Flush(...)` would be backward compatible (because if you don't look that response, it's not going to hurt to return it.

That would mean that both methods both `emit` and `return` the same `error`. I get that might not be what you are going for.

👍

tracer_helpers.go
MatthewDolan: Same comment as above, LGTM.

tedsuo:
Good catch @MatthewDolan .

So, It's painful to add two new methods, `CloseContext` and `FlushContext`, and deprecate `Close` and `Flush`. I'd rather rip the band-aid off, if we can gauge the impact to be minimal. 

Personally, I don't believe customers use these methods directly. However, assuming they do, but the cases are few, are there any mitigating steps we could take which would create less code debt than forever carting around deprecated methods on our brand new Tracer?

For example:
- Preserve the existing type as `type Tracerv0_14 interface` 
- Create a `NewTracerv0_14(Options) Tracerv0_14` function that wraps our new tracer 
- Could go 1.9 type aliasing help further mitigate this issue?

tedsuo:
Thinking more... I don't want deprecated methods... but given that it is straight forwards to implement these, and researching an alternative would add time to the PR, I can throw in the towel  on trying to keep the interface clean. There's enough coming down the pipe with the Go tracer that I don't want this PR to remain outstanding any longer.

MatthewDolan:
LGTM. I hear what you are saying about clients likely not having used these APIs. I approved the PR so feel free to merge with what you think is the best answer.

tedsuo:
In the end, I decided to remove old the methods from the Tracer interface. The determination was that the presence of error signatures, combined with concise method names, would perpetuate the bad usage patterns we are trying to eliminate. 

I added a deprecated `Tracerv0_14` interface, which works with the existing helper functions. If users were interacting with the old interface directly, they need to make the following changes:

- change the type: `Tracer` -> `Tracerv0_14`
- change the constructor `NewTracer(opts)` -> `NewTracerv0_14(opts)`

Users who are interacting with the tracer via the helper functions will not need to make any changes.

The expectation is that the compatibility issues are minor (at most, one or two call sites per service), so the cost of support is worth preventing the opportunity to block on close and flush.

PR #120, opened 2017-10-17
LS-1667 Split translation and Report processes in Go Client lib apart

aliceadelef:
This is the pre-requisite step needed to be made before we could add in timing and and count metrics to the go client.

### DIFF
--- a/grpc_collector_client.go
+++ b/grpc_collector_client.go
@@ -194,13 +194,23 @@ func (client *grpcCollectorClient) makeReportRequest(buffer *reportBuffer) *cpb.
 
 }
 
-func (client *grpcCollectorClient) Report(ctx context.Context, buffer *reportBuffer) (collectorResponse, error) {
-	resp, err := client.grpcClient.Report(ctx, client.makeReportRequest(buffer))
+func (client *grpcCollectorClient) Report(ctx context.Context, req *reportRequest) (collectorResponse, error) {
+	if req.grpcRequest == nil {
+		return nil, errors.InvalidArgument("grpcRequest cannot be null")
+	}
+	resp, err := client.grpcClient.Report(ctx, req.grpcRequest)
 	if err != nil {
 		return nil, err
 	}
 
 	return resp, nil
+}
+
+func (client *grpcCollectorClient) Translate(ctx context.Context, buffer *reportBuffer) reportRequest {
+	req := client.makeReportRequest(buffer)
+	return reportRequest{
+		grpcRequest: req,
+	}
 }
 
 func translateAttributes(atts map[string]string) []*cpb.KeyValue {
--- a/grpc_collector_client.go
+++ b/grpc_collector_client.go
@@ -14,6 +14,7 @@ import (
 	// N.B.(jmacd): Do not use google.golang.org/glog in this package.
 
 	google_protobuf "github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/lightstep/common-go/errors"
 	cpb "github.com/lightstep/lightstep-tracer-go/collectorpb"
 	ot "github.com/opentracing/opentracing-go"
 )
@@ -91,7 +92,7 @@ func (client *grpcCollectorClient) ConnectClient() (Connection, error) {
 
 		grpcClient, ok := uncheckedClient.(cpb.CollectorServiceClient)
 		if !ok {
-			return nil, fmt.Errorf("Grpc connector factory did not provide valid client!")
+			return nil, fmt.Errorf("grpc connector factory did not provide valid client")
 		}
 
 		conn = transport
@@ -194,7 +195,7 @@ func (client *grpcCollectorClient) makeReportRequest(buffer *reportBuffer) *cpb.
 
 }
 
-func (client *grpcCollectorClient) Report(ctx context.Context, req *reportRequest) (collectorResponse, error) {
+func (client *grpcCollectorClient) Report(ctx context.Context, req reportRequest) (collectorResponse, error) {
 	if req.grpcRequest == nil {
 		return nil, errors.InvalidArgument("grpcRequest cannot be null")
 	}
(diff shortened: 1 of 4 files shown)


grpc_collector_client.go
```diff
@@ -91,7 +91,7 @@ func (client *grpcCollectorClient) ConnectClient() (Connection, error) {
 
 		grpcClient, ok := uncheckedClient.(cpb.CollectorServiceClient)
 		if !ok {
-			return nil, fmt.Errorf("Grpc connector factory did not provide valid client!")
+			return nil, fmt.Errorf("grpc connector factory did not provide valid client")
```
MatthewDolan: nit: I have noticed that every error message inside of the tracer uses capital an exclamation. I definitely prefer the way you changed it, but it feels like it would be better to have a separate PR that changes all the errors at the same time.

grpc_collector_client.go
```diff
@@ -194,15 +194,25 @@ func (client *grpcCollectorClient) makeReportRequest(buffer *reportBuffer) *cpb.
 
 }
 
-func (client *grpcCollectorClient) Report(ctx context.Context, buffer *reportBuffer) (collectorResponse, error) {
-	resp, err := client.grpcClient.Report(ctx, client.makeReportRequest(buffer))
+func (client *grpcCollectorClient) Report(ctx context.Context, req reportRequest) (collectorResponse, error) {
+	if req.grpcRequest == nil {
```
MatthewDolan: Oh I miss generics so much :(

Instead of having a struct with many different types of pointers, what about replacing `reportRequest` with `interface{}` and then casting? I feel like that might be a little bit cleaner.

It's just a thought, I don't feel strongly about it.

interfaces.go
```diff
@@ -21,9 +23,15 @@ type collectorResponse interface {
 	Disable() bool
 }
 
+type reportRequest struct {
+	thriftRequest *lightstep_thrift.ReportRequest
+	grpcRequest   *cpb.ReportRequest
```
MatthewDolan: nit: this is actually a `protoRequest` not a `grpcRequest`. In the `v0.15.0` branch, I am adding proto over http and it uses the same request type.

interfaces.go
```diff
@@ -21,9 +23,15 @@ type collectorResponse interface {
...
+}
+
 // collectorClient encapsulates internal thrift/grpc transports.
 type collectorClient interface {
-	Report(context.Context, *reportBuffer) (collectorResponse, error)
+	Translate(context.Context, *reportBuffer) (reportRequest, error)
```
MatthewDolan: Thought: Could we separate out this translate code into an `Encoder` interface instead? I created a new `httpCollectorClient` in `v0.15.0` and it would be nice to share the `Encoder` between the two `clients`.

grpc_collector_client.go
jmacd: Yeah, I very nearly said the same thing, but don't feel strongly about it.

aliceadelef: huh @jmacd I could have sworn I ended up with the struct instead of an interface after a chat with you. Happy to do whatever we think makes more sense.

aliceadelef:
@MatthewDolan hmm yeah lightstep-tracer-common update was def not intentional.

jmacd:
Ah. Well, I was maybe confused, and I truly meant that I thought of saying
that and didn't because sometimes the answer is "Go fast".
ᐧ

grpc_collector_client.go
aliceadelef: @jmacd I could have sworn that I ended up chosing this approach over the interface after a conversation with you, but if that is not correct and neither of us can rembmer / come up with a good reason for why this appraoch was taken, then I am all for switching to interfaces. Just want to make sure there wasn't a good reason that I'm forgetting.

interfaces.go
aliceadelef: @MatthewDolan are you saying just to rename it to Encoder? or are you trying to get at something bigger here?

MatthewDolan: What I think I am trying to say is instead of adding `Translate(...)` to the `collectorClient` interface. Instead add a new interface:

```
type encoder interface {
  Translate(context.Context, *reportBuffer) (reportRequest, error)
}
```

There would be two encoders today (`protoEncoder` and `thriftEncoder`) and there would be three clients (`httpCollectorClient`, `grpcCollectorClient`, and `thriftCollectorClient`). The http and grpc implementations would share the same encoder.

Does that make sense?

aliceadelef: @MatthewDolan were you thinking of having the converter moved up a level then as well? right now the converter for PROTO lives in the client and in instantiated when the client is. I just want to make sure I'm understanding you correctly. You are proposing that the tracer call encoder.Translate and the encoder is is a part of the tracer and not the client? I know that you made some changes in this area recently so I want to make sure I don't alter something that you've already planned and thought about.

grpc_collector_client.go
aliceadelef: I think I'm going to leave this as is, since the actual requests are all specific the the networks and are not LS code, so making an interface would basically just be a wrapper. ie 
`type reportRequest interface {
      TypedRequest{} interface{}
}`
which i'm not certain how much cleaner that really makes things. Thinking I'll leave as is for now, especially since it sounds like much of this going to change soon anyway.

interfaces.go
MatthewDolan: We could definitely move it up into the tracer itself. I only put it in the client because I was trying to do the minimum change. I don't have strong feelings about where it lives.

grpc_collector_client.go
MatthewDolan: Totally ok with the decision to leave it as is, but just wanted to be clear, because I think we might be talking about a slightly different proposal.

I was suggesting that the function read like this:

```
func (client *grpcCollectorClient) Report(
   ctx context.Context,
   req interface{},
) (collectorResponse, error) {
  protoReq, ok := req.(*cp.ReportRequest)
  if !ok {
    return nil, fmt.Errorf("client / encoder combination not supported.")
  }

  ...
}
```

PR #121, opened 2017-10-19
Unify grpc and thrift transport tests.

MatthewDolan:


tracer_test.go
```diff
@@ -174,45 +174,46 @@ var _ = Describe("SpanRecorder", func() {
...
+				})
 			})
-		})
 
-		AfterEach(func() {
-			closeTestTracer(tracer)
-		})
+			AfterEach(func() {
+				closeTestTracer(tracer)
+			})
```
danielamiao: I find these sorts of "spacing" changes suspicious, just make sure `go fmt` is run? It could just be weird formatting before and you are actually fixing it now.

PR #122, opened 2017-10-21
LS-1667 Alice/client metrics/split report v0.15

aliceadelef:


proto_converter.go
```diff
@@ -113,13 +113,13 @@ func (converter *protoConverter) toField(key string, value interface{}) *cpb.Key
...
 		field.Value = &cpb.KeyValue_BoolValue{BoolValue: reflectedValue.Bool()}
 	default:
-		field.Value = &cpb.KeyValue_StringValue{StringValue: fmt.Sprint(reflectedValue)}
-		maybeLogInfof(
-			"value: %v, %T, is an unsupported type, and has been converted to string",
-			converter.verbose,
-			reflectedValue,
-			reflectedValue,
-		)
+		var s string
```
aliceadelef: due to the funky branch merging going on, this is a change I pulled in from a fix Josh made recently. https://github.com/lightstep/lightstep-tracer-go/pull/114

interfaces.go
```diff
@@ -21,9 +24,16 @@ type collectorResponse interface {
...
+	thriftRequest *lightstep_thrift.ReportRequest
+	protoRequest  *cpb.ReportRequest
+	httpRequest   *http.Request
+}
+
 // collectorClient encapsulates internal thrift/grpc transports.
 type collectorClient interface {
-	Report(context.Context, *reportBuffer) (collectorResponse, error)
+	Report(context.Context, reportRequest) (collectorResponse, error)
+	Translate(context.Context, *reportBuffer) (reportRequest, error)
```
MatthewDolan: I think in the rebase my comments disappeared. I feel like it would be nice to split this into `collectorClient` and `collectorEncoder` (then we can unify the code across `grpcCollectorClient` and `httpCollectorClient` into `protoCollectorEncoder` and split `thriftCollectorClient` into the client and a new `thriftCollectorEncoder` or something like that.

Although the code definitely is pretty readable as is, so I don't think it's a big deal to leave it this way.

Mostly I am just thinking this is good because I know while brainstorming it was suggested that down the line we might want to have a more clear separation between an unencoded queue and an encoded queue and having these two things split would help that.

MatthewDolan: Chatted offline. LGTM!

proto_converter.go
jmacd: :git_confusion:

PR #124, opened 2017-10-27
Support error values in proto

tedsuo:
For unknown objects, we support `fmt.Stringer`. We should probably support `error` as well. Not sure if there any other common serializable interfaces we should check for.

- Typecast unknown objects to error and encode the error string.
- do not emit `Unsupported Value` event.

### DIFF
--- a/proto_converter.go
+++ b/proto_converter.go
@@ -115,9 +115,12 @@ func (converter *protoConverter) toField(key string, value interface{}) *cpb.Key
 		field.Value = &cpb.KeyValue_BoolValue{BoolValue: reflectedValue.Bool()}
 	default:
 		var s string
-		if stringer, ok := value.(fmt.Stringer); ok {
-			s = stringer.String()
-		} else {
+		switch value := value.(type) {
+		case fmt.Stringer:
+			s = value.String()
+		case error:
+			s = value.Error()
+		default:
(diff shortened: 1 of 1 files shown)


proto_converter.go
```diff
@@ -115,9 +115,12 @@ func (converter *protoConverter) toField(key string, value interface{}) *cpb.Key
 		field.Value = &cpb.KeyValue_BoolValue{BoolValue: reflectedValue.Bool()}
 	default:
 		var s string
-		if stringer, ok := value.(fmt.Stringer); ok {
-			s = stringer.String()
-		} else {
+		switch value := value.(type) {
+		case fmt.Stringer:
+			s = value.String()
+		case error:
```
MatthewDolan: Do some `error`s implement `Stringer`? I feel like you might want to reverse the order of these two. But also that's assuming that switch executes in order, which I don't know is guaranteed.

jmacd:
They do execute in order.

proto_converter.go
tedsuo: Thankfully, `switch` always executes in order. When an error also matches `fmt.Stringer`, the intent is usually to allow the error to print the error message correctly when it's handed to `fmt.Print()`. Since `fmt.Stringer` is specifically for printing, it's probably okay as a default. But my guess is the order almost never matters.

PR #127, opened 2017-10-27
Add EventTracerDisabled

tedsuo:
Missed this one.

- emit a EventTracerDisabled rather than printing to screen.
- but don't emit the event inside of the lock.

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -314,16 +314,16 @@ func (tracer *tracerImpl) postFlush(flushEventError *eventFlushError) *eventStat
 
 func (tracer *tracerImpl) Disable() {
 	tracer.lock.Lock()
-	defer tracer.lock.Unlock()
+	disabled := tracer.disabled
+	tracer.disabled = true
+	tracer.buffer.clear()
+	tracer.lock.Unlock()
 
-	if tracer.disabled {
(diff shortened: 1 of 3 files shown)


tracer.go
```diff
@@ -314,16 +314,16 @@ func (tracer *tracerImpl) postFlush(flushEventError *eventFlushError) *eventStat
 
 func (tracer *tracerImpl) Disable() {
 	tracer.lock.Lock()
-	defer tracer.lock.Unlock()
+	disabled := tracer.disabled
```
MatthewDolan: It seems like you are changing behavior here (i.e. disabling a disabled tracer does something different).

I assume you are trying to get around emitting the event within the scope of the lock. What about something like this?
```
var disabled bool
defer func() {
  if !disabled {
    emitEvent(newEventTracerDisabled())
  }()
  tracer.lock.Lock()
  defer tracer.lock.Unlock()

  disabled = tracer.disabled

  if disabled {
    return
  }

  tracer.buffer.clear()
  tracer.disabled = true
}
```

tedsuo: The behavior change is effectively a non-op, but I'm fine with keeping it the same. Above LGTM.

PR #135, opened 2017-11-27
Avoid reflection with %T formatter

jmacd:


jmacd:
Are we going to start a 0.16 branch?

jmacd:
Should we use branches or master?

MatthewDolan:
Um, I feel like we should cut another branch again (if no one has any objections). That process worked well and helped keep these changes off of master where they could be picked up with `go get`. I just named it whatever the version I expect it to be released under.

jmacd:
I agree. I find working on a branch leads to more git mistakes. That's all.

PR #139, opened 2017-12-08
Add Gopkg setup

jmacd:


jmacd:
I had trouble getting `dep` to work from our repos without this.

PR #141, opened 2018-01-16
Add GetLightStepReporterID helper

jmacd:


jmacd:
Something has to be done, since the SpanRecorded API doesn't expose the variable I need. This is to allow upgrading our `sentinel` to 0.15.x, it's running an ancient version.

tedsuo:
Seems like a fine feature. I wish these functions didn’t start with “Get” as that is against go coding standards, but whatever. 

I know it’s kind of pedantic, but there should be a test for each case: tracer, tracer_v14, and unsupported tracer.

PR #148, opened 2018-07-06
support io.Writer/io.Reader for Inject/Extract

### LINKED ISSUE #147: Binary propagator does not accept the io.Reader/io.Writer
The `github.com/opentracing/opentracing-go` says:
```go
const (
    // Binary represents SpanContexts as opaque binary data.
    //
    // For Tracer.Inject(): the carrier must be an `io.Writer`.
    //
    // For Tracer.Extract(): the carrier must be an `io.Reader`.
    Binary BuiltinFormat = iota
```
calling `sc, err := ot.GlobalTracer().Extract(ot.Binary, bytes.NewBuffer(traceData))` always results in `ot.ErrInvalidCarrier`

vetinari:
This supports io.Writer for tracer.Inject() and io.Reader for
tracer.Extract() as required by github.com/opentracing/opentracing-go.

Also: fixes to accept opentracing.Binary as parameter for Inject/Extract
instead of lightstep.BinaryCarrier

fixes #147

### DIFF
--- a/propagation_binary.go
+++ b/propagation_binary.go
@@ -11,7 +11,7 @@ import (
 )
 
 // BinaryCarrier is used as the format parameter in inject/extract for lighstep binary propagation.
-const BinaryCarrier = "lightstep_binary_carrier"
+const BinaryCarrier = opentracing.Binary
 
 var theBinaryPropagator binaryPropagator
 
@@ -36,14 +36,13 @@ func (binaryPropagator) Inject(
 	if err != nil {
 		return err
 	}
-	if carrier, ok := opaqueCarrier.(io.Writer); ok {
+
+	switch carrier := opaqueCarrier.(type) {
+	case io.Writer:
 		buf := make([]byte, base64.StdEncoding.EncodedLen(len(data)))
 		base64.StdEncoding.Encode(buf, data)
 		_, err = carrier.Write(buf)
 		return err
-	}
-
-	switch carrier := opaqueCarrier.(type) {
 	case *string:
 		*carrier = base64.StdEncoding.EncodeToString(data)
 	case *[]byte:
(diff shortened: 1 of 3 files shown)


propagation_binary.go
```diff
@@ -34,6 +36,13 @@ func (binaryPropagator) Inject(
 	if err != nil {
 		return err
 	}
+	if carrier, ok := opaqueCarrier.(io.Writer); ok {
+		buf := make([]byte, base64.StdEncoding.EncodedLen(len(data)))
+		base64.StdEncoding.Encode(buf, data)
```
MatthewDolan: Mind adding a comment here to how you determined use `base64.StdEncoding`? I know that this as to inter-op with the tracers in other languages and I assume that they just all use base64 but it would be nice to have a reference to that decision somewhere.

tracer.go
```diff
@@ -139,7 +139,7 @@ func (tracer *tracerImpl) Inject(sc ot.SpanContext, format interface{}, carrier
 	switch format {
 	case ot.TextMap, ot.HTTPHeaders:
 		return theTextMapPropagator.Inject(sc, carrier)
-	case BinaryCarrier:
```
MatthewDolan: To make this backward compatible, can we make `const BinaryCarrier = ot.Binary` at `propagation_binary.go:12` too?

propagation_binary.go
vetinari: I just copied that from the other string / []byte cases...

MatthewDolan: Ah, I see. I will do some more digging on my end. Please ignore the above comment. No action is needed.

joeblubaugh:
CircleCI is being painful with this change - it's not running tests after I updated a setting that should cause them to be run. I opened #150, which should get around this. Thanks for the fix!

joeblubaugh:
And, when I created the new PR it also kicked off this build. Hm.

PR #152, opened 2018-07-23
LS-4640 Widen timestamp_offset to 64 bits

jmacd:


jmacd:
This is practically a no-op but it reminds me that `lightstep-tracer-go` is not using the code generated by the Makefile in `lightstep-tracer-common`, and there are issues to resolve because that code generates both Gogo and Protobuf code. Could we / should we support more than one transport library? Ughh.

PR #154, opened 2018-08-03
add random pool to reduce concurrency contention on genSeededGUID()

lzuwei:
- genSeededGUID() has a potential to content on a single mutex lock with high parallelism as observed in our production system at Grab
- mitigate this issue using a pool of random generators wrapped with a lock to reduce contention

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,24 @@
+version: 2
+
+jobs:
+  test:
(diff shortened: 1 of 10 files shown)


jmacd:
The interface should have two methods,
`func Int63() int64` and `func TwoInt63() (int64, int64)`
or something like that.

lzuwei:
@jmacd @tedsuo, I have shifted the implementation details into the suggested packages. Please review, thank you!

PR #160, opened 2018-09-07
fix http double headers

vetinari:
fix double headers when reusing the *http.Request ￼…
because of https://github.com/opentracing/opentracing-go/issues/159
the headers in a reused *http.Request are appended instead of set
which results in headers like this after injecting the current span
context:
```
ot-tracer-spanid: 911775291efa49de,46fbd688b66dc96d
ot-tracer-traceid: 24080003fd414b3d,24080003fd414b3d
ot-tracer-sampled: true,true
```

With these double headers at least the java client is unable to
extract the current span context.


vetinari:
see also https://github.com/opentracing/opentracing-go/pull/191

joeblubaugh:
@vetinari thanks for the PR. Would you also like to file an issue against the Java repository? It's an open question to me how we should handle a list in the header format, but crashing definitely isn't it.

MatthewDolan:
I went back and realized that we do have a Gopkg.lock file. We should update that file to go beyond opentracing/opentracing-go#191, let me dig in deeper here.

PR #166, opened 2018-10-03
Fixes null pointer exception when using HTTP client

bg451:
There is currently an issue in the HTTP client where a null httpRequest causes a panic. The cause for the null request in the first place is unknown, so the goal for this PR is to mitigate the panic and correctly capture errors returned by  `(client).Translate`. 

- [x] mitigate null pointer exception
- [] add test to test behavior of translate errors and null requests
- [] discover root cause of the http translation error

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -231,8 +231,12 @@ func (tracer *tracerImpl) Flush(ctx context.Context) {
 	defer cancel()
 
 	req, flushErr := tracer.client.Translate(ctx, &tracer.flushing)
-	resp, flushErr := tracer.client.Report(ctx, req)
+	if flushErr != nil {
+		emitEvent(newEventFlushError(flushErr, FlushErrorTranslate))
+		return
+	}
 
+	resp, flushErr := tracer.client.Report(ctx, req)
 	if flushErr != nil {
 		flushErrorEvent = newEventFlushError(flushErr, FlushErrorTransport)
 	} else if len(resp.GetErrors()) > 0 {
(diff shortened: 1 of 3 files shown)


bg451:
It's a little unfortunate that there's a little bit of code duplication from the `test_utils_test.go`, perhaps there's some work we can do to move a lot of the mocks into a subpackage.

tracer.go
```diff
@@ -219,35 +219,38 @@ func (tracer *tracerImpl) RecordSpan(raw RawSpan) {
...
+		errorEvent := newEventFlushError(err, FlushErrorTranslate)
+		emitEvent(errorEvent)
+		// call postflush to prevent the tracer from going into an invalid state.
+		emitEvent(tracer.postFlush(errorEvent))
+		return
+	}
 
-	if flushErr != nil {
-		flushErrorEvent = newEventFlushError(flushErr, FlushErrorTransport)
+	var errorEvent *eventFlushError
```
joeblubaugh: what's up here with the multiple variables named `errorEvent` in the same fn?

bg451:
@joeblubaugh in the interest of getting a patch out to a customer, I'm going to go ahead and merge this. @JulianGriggs and I walked through this code and there's a lot of refactor work that needs to be done in order to get the Flush method in abetter state. I have time carved out at the EOD to deep dive into the code. 

The issue comes down to postFlush being fairly complex. It'd be nice to the decouple the piece for going to a not flushing state, creating the status report, and determining whether to retry/clear the buffer.

PR #167, opened 2018-10-04
Update SendSpan with options to send via different transports

JulianGriggs:
Minor update to SendSpan to make it easier to pick which transport protocol to use


cmd/sendspan/main.go
```diff
@@ -1,46 +1,61 @@
...
 )
 
 var (
 	flagAccessToken = flag.String("access_token", "", "Access token to use when reporting spans")
-
-	flagHost   = flag.String("collector_host", "", "Hostname of the collector to which reports should be sent")
-	flagPort   = flag.Int("collector_port", 0, "Port of the collector to which reports should be sent")
-	flagSecure = flag.Bool("secure", true, "Whether or not to use TLS")
-
-	flagUseGRPC = flag.Bool("use_grpc", true, "Whether or not to use gRPC")
```
bg451: Since this is a potentially breaking configuration change, can we continue to support it but mark it as deprecated?

JulianGriggs: I'd care more if we were at 1.x instead of 0.x

bg451: Oops for some reason I thought I was looking at a different file, did not realize this was send span.

PR #172, opened 2018-10-09
run dep ensure

MatthewDolan:
R: @jmacd @JulianGriggs 

### Summary of changes

- Ran `dep ensure` with v0.5.0 (https://github.com/golang/dep/releases/tag/v0.5.0) of dep.
- Updated prune options to remove unused packages and tests.
- vendor-ed in dependencies and committed them to the repo.


JulianGriggs:
Should we document which version of dep we are using? A lot of these changes look like they are the result of a dep upgrade.

MatthewDolan:
@JulianGriggs Good point. I added it to the description.

JulianGriggs:
Maybe we can add it to the README as well? Also, do you have any idea why we didn't have these vendored in already?

MatthewDolan:
I don't know what the context is behind us not vendor-ing.

I have seen quite a few public repos quoting the Go spec.

> Go 1.6 includes support for using local copies of external dependencies to satisfy imports of those dependencies, often referred to as vendoring.

which suggests we should vendor if we want to allow people to `go get` our code.

MatthewDolan:
Closing for now, it might make the code a little better but this isn't solving a direct issue.

Issue #149, opened 2018-07-07
Importing lightstep-tracer-go bloats programs and significantly slows down build time


achille-roussel:
Hello, I've recently integrated a program with lightstep using the lightstep-tracer-go package and the size of the compiled binary went from 17MB to 24MB and build time increased by over 2x. It seems like the package has lots of dependencies, to give an idea of the situation:
```
# in lightstep-tracing-go, listing external dependencies
$ govendor list +e | wc -l
      69
```
Some of those dependencies are sub-packages of grpc and others but still... it's quite a lot.

What's most problematic about this is a lot of this code is unused in the resulting application (because the program will make use of either grpc, thrift, or http to send spans to the collectors), but the compiler cannot eliminate this code since it's not _dead code_, because the selection of the protocol happens at runtime through the options passed to the tracer when it's created.

Here's an example showing the cost of integrating lightstep-tracer-go in one of or production services (measurements done on OSX 17.5.0 with Go 1.10).
```
# without lightstep-tracer-go
$ time go build ./cmd/main/

real	0m2.540s
user	0m3.177s
sys	0m0.549s
$ ls -l main
-rwxr-xr-x 1 achille achille 17828770 Jul  7 00:23 main
```
```
# with lightstep-tracer-go
$ time go build ./cmd/main/

real	0m5.854s
user	0m1.911s
sys	0m4.510s
$ ls -l main
-rwxr-xr-x 1 achille achille 24043642 Jul  7 00:23 main
```
The main problem with slowing down build time is it gets in the way of developer productivity, more than doubling is a significant cost that we pay when attempting to iterate quickly on a project. The example I'm giving you is from a project that has already quite a bit of code, to put things in perspective here is the number of lines there were before adding lightstep-tracer-go (but including all other dependencies):
```
$ loc -u
--------------------------------------------------------------------------------
 Language             Files        Lines        Blank      Comment         Code
--------------------------------------------------------------------------------
 Go                     677       256832        25115        20994       210723
 Markdown                19         2439          537            0         1902
 Perl                     8         1141          142          140          859
 Assembly                23          877          217           24          636
 Bourne Shell             3          787           65          260          462
 JSON                     1          255            0            0          255
 Python                   2          128           24            6           98
 YAML                     4           83           11            6           66
 Makefile                 2           48           11            0           37
 C                        1           47           11            7           29
 JavaScript               2           16            4            0           12
--------------------------------------------------------------------------------
[...]

PR #174, opened 2018-10-22
use standard context package instead of x/net/context

achille-roussel:
Following up on our conversation in https://github.com/lightstep/lightstep-tracer-go/issues/149

This PR modifies the source files to use the standard `context` package instead of `x/net/context`. This means that this repository will not be compatible with Go 1.6 and earlier anymore. The README does not mention the minimum Go version currently supported so I'm not sure if this is going to be a deal-breaker or not.

If we need support for Go 1.6 we can probably figure out a way to support either packages with a compile-time check, but it could introduce lots of complexity in the code.

I was concerned that `x/net/http2` may have used `x/net/context` directly but it seems they have moved to the standard `context` package already, there's only one leftover in a test:
```
github.com/golang/net/http2 $ grep -r 'x/net/context' .
./transport_test.go:	"golang.org/x/net/context"
```

There are two places where `x/net/context` is still referenced after this change:
```
lightstep-tracer-go $ grep -r x/net/context .
./collectorpb/collectorpbfakes/fake_collector_service_client.go:	"golang.org/x/net/context"
./collectorpb/collector.pb.go:	context "golang.org/x/net/context"
```
I tried re-generating those files but they kept importing `x/net/context`, I'm not super familiar with the build process here so I could take a pointer or two on how to fix this.


iredelmeier:
LGTM. We'll deal with the `x/net/context` in the fakes separately.

PR #175, opened 2018-10-22
remove x/net/http2 dependency

achille-roussel:
This PR is a follow up from our conversation in https://github.com/lightstep/lightstep-tracer-go/issues/149

The http client implementation was using `http.Transport` for plain http, and `http2.Transport` for https, which is already the default behavior of the standard library (since http2 support was introduced in Go 1.6).

It seems like the extra dependency on `x/net/http2` is unnecessary and the lightstep client could use the default http transport without losing any features.

The PR also fixes support for http proxies, and adds a couple of sane defaults for various configuration options of the http transport.


collector_client_http.go
```diff
@@ -71,37 +71,43 @@ func newHttpCollectorClient(
...
-	// For now, we are falling back to http1 for plaintext.
-	// In the future, we might want to add out own h2c implementation (see https://github.com/hkwi/h2c).
-	var transport http.RoundTripper
-	if client.url.Scheme == "https" {
-		transport = &http2.Transport{}
-	} else {
-		transport = &http.Transport{}
+	// Use a transport independent from http.DefaultTransport to provide sane
+	// defaults that make sense in the context of the lightstep client.
+	transport := &http.Transport{
```
iredelmeier: Where they differ significantly from `DefaultTransport`, how did you choose parameters? e.g., are you experiencing an issue fixed by `DisableCompression: true`?

achille-roussel: `DisableCompression` here only controls compression of the response body, but the client is mostly doing uploads and the collectors aren't returning anything in the response, so it's a no-op.
This is just a setting to indicate explicitly that there's no point using compression, I can add a comment if you want.

The thing is, the zero-value of `http.Transport` is different from `http.DefaultTransport`, so the previous code already had a different behavior (and actually wasn't working with HTTP proxies), here's the standard configuration:
```
var DefaultTransport RoundTripper = &Transport{
        Proxy: ProxyFromEnvironment,
        DialContext: (&net.Dialer{
                Timeout:   30 * time.Second,
                KeepAlive: 30 * time.Second,
                DualStack: true,
        }).DialContext,
        MaxIdleConns:          100,
        IdleConnTimeout:       90 * time.Second,
        TLSHandshakeTimeout:   10 * time.Second,
        ExpectContinueTimeout: 1 * time.Second,
}
```
https://golang.org/pkg/net/http/#RoundTripper

iredelmeier: Yes, I see that the behaviour is different. The context of my question had been the reference to `DefaultTransport` in the context, since it seemed like the proposed `Transport` was meant to be juxtaposed against `DefaultTransport` rather than `http.Transport{}`.

Do you mind adding that comment about `DisableCompression`?

I definitely agree re: adding timeouts, etc. 👍

achille-roussel: Done!

PR #183, opened 2018-10-30
Allow new spans to have external references

### LINKED ISSUE #182: newSpan not confirming type assertion (panic)
Is this intentional? This line specifically [here](https://github.com/lightstep/lightstep-tracer-go/blob/master/span.go#L50) should probably be `refCtx, ok` and if it's `!ok`, it should bail out. Unless there's a good reason for it panicking that I'm unaware of. I was burned by this today unexpectedly, so any insights would be great. Thank you!

iredelmeier:
[Fixes #182]

### DIFF
--- a/tracer_test.go
+++ b/tracer_test.go
@@ -106,7 +106,7 @@ var _ = Describe("Tracer", func() {
 			}).To(Not(Panic()))
 		})
 
-		It("can start a spn withreferences to external spans", func() {
+		It("can start a span with references to external spans", func() {
 			otherTracer := opentracing.NoopTracer{}
 			otherSpan := otherTracer.StartSpan("other")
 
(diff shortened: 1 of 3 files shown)


tracer_test.go
```diff
@@ -95,6 +95,27 @@ var _ = Describe("Tracer", func() {
...
+			otherSpan := tracer.StartSpan("other")
+
+			Expect(func() {
+				tracer.StartSpan("test", opentracing.SpanReference{
+					ReferencedContext: otherSpan.Context(),
+				})
+			}).To(Not(Panic()))
+		})
+
+		It("can start a spn withreferences to external spans", func() {
```
joeblubaugh: s/withreferences/with references/

joeblubaugh: s/spn/span

PR #185, opened 2019-01-04
Adds conformance test LS-6254

bg451:
We have implemented a conformance test which validates a tracer's inject/extract carrier format. There is a conformance harness that sends a set of carriers to a client (written in python here) and reads the clients response from stdout.

The client reads a JSON blob from stdin, extracts the span context for each carrier format, then injects the span contexts into a new JSON blob and sends that over stdout.

To test this out locally, install the local build runner viagcloud components install cloud-build-local and then run cloud-build-local --dryrun=false --config=cloudbuild.yaml .


kayousterhout:
Closing this due to inactivity

PR #187, opened 2019-01-28
Add meta event reporting LS-6841

austinlparker:
This PR adds support for reporting 'meta events' - detailed lifecycle information about a tracer and its spans - to the Go tracer. It can be enabled via the `MetaEventLogging` flag in options. This option is not suitable for production environments as it generates a significant amount of spans.


austinlparker:
Big note: This PR also removes all thrift support from the library.

austinlparker:
@MatthewDolan We could, but we don't implement the DevMode flag in Thrift so this one would be kinda hacky.

iredelmeier:
Yeah, I agree with @MatthewDolan that it feels like Thrift should be removed independently. I can open a separate PR to do that as a first step.

austinlparker:
Nah, I can open one real quick @iredelmeier

austinlparker:
@iredelmeier @MatthewDolan see #189

austinlparker:
@iredelmeier @MatthewDolan @jmacd The separate thrift PR has been merged and this one is clean for review.

span.go
```diff
@@ -77,6 +77,14 @@ ReferencesLoop:
 	sp.raw.Start = startTime
 	sp.raw.Duration = -1
 	sp.raw.Tags = opts.Options.Tags
+
+	if tracer.opts.MetaEventLogging && IsMetaSpan(sp) {
```
jmacd: Should we include the starting tags in the meta event? I'm imagining that if we were filtering a view and wanted to find unfinished spans, it would be helpful to know the tags of the unfinished span.

span.go
```diff
@@ -233,6 +241,13 @@ func (s *spanImpl) FinishWithOptions(opts ot.FinishOptions) {
 	s.raw.Duration = duration
 
 	s.tracer.RecordSpan(s.raw)
+	if s.tracer.opts.MetaEventLogging && IsMetaSpan(s) {
+		ot.StartSpan(LSMetaEvent_SpanFinishOperation,
+			ot.Tag{Key: LSMetaEvent_MetaEventKey, Value: true},
```
jmacd: ^^^ about Finish events.

meta_events.go
```diff
@@ -0,0 +1,15 @@
...
+const LSMetaEvent_PropagationFormatKey = "lightstep.propagation_format"
+const LSMetaEvent_TraceIdKey = "lightstep.trace_id"
+const LSMetaEvent_SpanIdKey = "lightstep.span_id"
+const LSMetaEvent_TracerGuidKey = "lightstep.tracer_guid"
+
+// operation names for ls meta span events
+const LSMetaEvent_ExtractOperation = "lightstep.extract_span"
+const LSMetaEvent_InjectOperation = "lightstep.inject_span"
+const LSMetaEvent_SpanStartOperation = "lightstep.span_start"
+const LSMetaEvent_SpanFinishOperation = "lightstep.span_finish"
```
jmacd: I don't see much value in a Finish meta event, since we just finished the actual span which includes all the same (and more) detail.

options.go
```diff
@@ -183,6 +183,9 @@ type Options struct {
 
 	// For testing purposes only
 	ConnFactory ConnectorFactory `yaml:"-" json:"-"`
+
+	// Enable LightStep Meta Event Logging
+	MetaEventLogging bool `yaml:"meta_event_logging" json:"meta_event_logging"`
```
jmacd: Making this an option sounds like leaving it unset amounts to opting out, i.e., this looks like an opt-in flag. It's not treated as an opt-out flag, however, it's mutated which feels wrong for an options struct. Should this be moved into a non-options struct?

util.go
```diff
@@ -30,3 +30,10 @@ func genSeededGUID2() (uint64, uint64) {
 	n1, n2 := randompool.Pick().TwoInt63()
 	return uint64(n1), uint64(n2)
 }
+
+func IsMetaSpan(s *spanImpl) bool {
+	if _, ok := s.raw.Tags["lightstep.meta_event"]; ok {
```
jmacd: Maybe take advantage of Go's default value semantics

```
return s.raw.Tags["lightstep.meta_event"] != nil
```

meta_events.go
austinlparker: My rationale with including start and finish is that you could easily determine how many unfinished spans exist via sets. It'd be a superset of spans, so you could determine "in-flight" spans more quickly?

options.go
austinlparker: Would it be more helpful to rename it? `MetaEventReportingEnabled`?

util.go
austinlparker: 👍 good call, thanks!

options.go
jmacd: I would say `MetaEventReportingDisabled`, that way the user gets the default on behavior via Go's struct initialization, and has to set `true` to opt-out of the feature.

But I wouldn't modify the value passed in by the user, have a field in the `tracerImpl` which is set to `!MetaEventReportingDisabled && DevModeEnabledBySatellite()`.

meta_events.go
jmacd: I can see some ways in which the analysis is more straightforward if we're simply subtracting finishes from starts. But, I feel that if we think about this in the form of relational queries, there's not much difference between selecting the finished span_ids from the stream of real spans vs selecting the finished span_ids from the stream of meta spans. https://lightstep.slack.com/archives/CFNR8UUNL/p1549575209243200

jmacd: I've reconsidered. Let's go ahead with a finish event.

options.go
jmacd: Shouldn't the satellite disable this feature except in dev-mode? It seems easier for the user to get started if this is on-by-default-as-long-as-the-satellite-says.

meta_events.go
jmacd: I was imagining a what-if scenario where we tried to forward these meta-spans to a separate lightstep instance, and to actually think of them as non-meta span data in this way. Then, I think, it would be nice if these meta spans formed meta-traces. This would require somehow coordinating so that all the meta-spans corresponding to a span have the same meta-trace_id. Can we do that?

austinlparker: For a single process, it'd be possible (but probably kinda slow) - the tracer would need a map of meta-contexts to their associated live spans that would be referenced for every finish/inject.

The bigger problem is mostly, imo, RPCs. You could theoretically have inject reference that meta-context map then inject the meta-context as well as the live context (we'd need a different, new, header for this - something like `x-lightstep-meta-s/span/trace-id`) so that the associated meta-start event on the callee wouldn't break.

My thought would be that it'd be faster and easier to reassemble them on the remote side if we really wanted a 'trace' view of the meta events.

options.go
austinlparker: The satellite either sends `devMode: true` or nothing, there's no associated `devMode: false`. If we wanted to add that, the tracers could check for it.

span.go
austinlparker: I think I'd like to hold off on increasing the size of these meta spans for now. I think we should keep it mind for the future though.

options.go
jmacd: I mean to presume `devMode: false` unless `devMode: true` is set. Then a default of `MetaEventReportingDisabled: false` would allow dev-mode satellites to enable meta-reporting clients out of the box, while production satellites would automatically leave meta-reporting disabled.

meta_events.go
jmacd: I didn't mean for the meta "trace" to cross span boundaries, just one meta-trace per span. The meta trace-id could be derived by hashing the span_id of the span. Then if we directed this traffic to a lightstep instance of our own, we'd have a bunch of traces representing every span being sent.

options.go
austinlparker: I pushed a change that does effectively this, if the devmode flag isn't explicitly there then it'll disable reporting.

meta_events.go
austinlparker: That'd be a pretty big change in scope since we'd need to have a separate code path to create the derivative trace id, then whatever hash function we're using would need to be ported across all the other tracer libraries.

PR #189, opened 2019-02-01
Remove Thrift

austinlparker:
Note - whitespace changes are a result of gofmt/goimports


jmacd:
I want to block this until we have a solution for load-testing Thrift at the satellite.

austinlparker:
@jmacd donno yet, but it'll be a major bump

PR #192, opened 2019-02-25
Switch from dep to go modules

iredelmeier:
Note: this is a breaking change.


bruceadowns:
Hi @iredelmeier (and @austinlparker),

I created a similar PR and looks like you may be missing a couple of pieces here. See https://github.com/lightstep/lightstep-tracer-go/pull/214

Tho my build is also failing, due to what seems to be an unrelated race condition.

iredelmeier:
Closing in favour of #214

PR #196, opened 2019-02-27
Add experimental OpenCensus exporter

iredelmeier:
Some open questions:
* Are all of the OC -> OT conversions appropriate? e.g., is this how we should handle `SpanData.Event`?
* What, if anything, should we do with the `MessageEvent`s? Similarly, how should we handle non-child `Link`s?
* What additional configuration, if any, do we want to expose?
* Should this be in a separate repo or this one?

TODO:
* Tests (!)
* README
* Documentation

### DIFF
--- a/lightstepoc/exporter.go
+++ b/lightstepoc/exporter.go
@@ -13,13 +13,18 @@ import (
 )
 
 var (
+	// ErrFailedToCreateExporter indicates that the underlying tracer could not be created,
+	// but the root reason is not known.
 	ErrFailedToCreateExporter = errors.New("lightstepoc: failed to create exporter")
 )
 
+// Exporter may be registered with OpenCensus so that span data can be exported to LightStep
 type Exporter struct {
 	tracer lightstep.Tracer
 }
 
+// NewExporter creates a new Exporter.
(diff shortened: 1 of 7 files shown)


lightstepoc/exporter.go
```diff
@@ -0,0 +1,104 @@
...
+
+func NewExporter(opts ...Option) (*Exporter, error) {
+	c := defaultConfig()
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	tracer := lightstep.NewTracer(c.tracerOptions)
+	if tracer == nil {
+		// Ideally we'd know *why* we couldn't create the tracer, but `NewTracer` currently doesn't return errors
```
etsangsplk: Lightstep Options Validate returns error https://github.com/lightstep/lightstep-tracer-go/blob/6254a61a022ea44a8e6dd828e06d40a4ea7bc0b3/options.go#L261
Which NewTraaer already takes care of, but you get the actual err by calling it yourself.

iredelmeier:
@austinlparker I'm not quite sure re: attributes -> tags, although I believe that all the types are compatible (i.e., that attributes support a subset of the types supported by tags).

PR #197, opened 2019-03-09
remove locally compiled protos with protos compiled in lightstep-tracer-common

MatthewDolan:
R: @iredelmeier @jmacd 

# Summary of Change

I noticed that we have two compiled version of the protos. I am doing some work to increment our gogo version and having a single version of the protos will help simplify that work.

What do you think of this change?


Gopkg.toml
```diff
@@ -17,3 +17,7 @@
 [[constraint]]
   name = "google.golang.org/grpc"
   version = "1.4.3"
+
+[[constraint]]
+  name = "github.com/lightstep/lightstep-tracer-common"
+  version = "1.0.2"
```
jmacd: Use 1.0.3.

PR #198, opened 2019-03-09
remove unused dep constraint

MatthewDolan:
R: @iredelmeier @jmacd

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -11,9 +11,5 @@
   version = "1.0.2"
 
 [[constraint]]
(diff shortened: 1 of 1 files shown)


jmacd:
I hope the newer protocs do not generate this x/net stuff.

MatthewDolan:
@jmacd this pr is in `lightstep-tracer-go`. The newer protoc files will be generated into `lightstep-tracer-common` so even if it does end up using x/net (which it hopefully will not) this repo won't need it as a dependency.

PR #200, opened 2019-03-11
Update to gogo 1.2.1

jmacd:
Use gogo 1.2.1 based on lightstep-tracer-common 1.0.3

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,14 +2,6 @@
 
 
 [[projects]]
-  digest = "1:dc62e76fd82a4bd369b92e8adc26d69d18c46d194f183179ac06c92fa4360472"
-  name = "github.com/Pallinder/go-randomdata"
-  packages = ["."]
-  pruneopts = ""
-  revision = "97a2356fcab20708fb8ae46cbbf69a5bc9feca63"
-  version = "v1.1.0"
-
-[[projects]]
   digest = "1:fd53b471edb4c28c7d297f617f4da0d33402755f58d6301e7ca1197ef0a90937"
   name = "github.com/gogo/protobuf"
   packages = [
@@ -223,7 +215,6 @@
   analyzer-name = "dep"
   analyzer-version = 1
   input-imports = [
-    "github.com/Pallinder/go-randomdata",
     "github.com/gogo/protobuf/proto",
     "github.com/gogo/protobuf/types",
     "github.com/lightstep/lightstep-tracer-common/golang/gogo/collectorpb",
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,26 @@
 
 
 [[projects]]
+  digest = "1:dc62e76fd82a4bd369b92e8adc26d69d18c46d194f183179ac06c92fa4360472"
+  name = "github.com/Pallinder/go-randomdata"
+  packages = ["."]
+  pruneopts = ""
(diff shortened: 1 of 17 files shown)


Gopkg.toml
```diff
@@ -1,3 +1,7 @@
+required = [
+    "github.com/gogo/protobuf/proto",
```
MatthewDolan: I don't think this should be necessary and I actually think adding it might result in extra files we don't want.

collector_client_http.go
```diff
@@ -10,8 +10,8 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/golang/protobuf/proto"
-	"github.com/lightstep/lightstep-tracer-common/golang/protobuf/collectorpb"
+	"github.com/gogo/protobuf/proto"
```
MatthewDolan: Now that we are importing gogo directly, we should add a `constraint` to the Gopkg.toml file on the version.

Gopkg.lock
```diff
@@ -1,6 +1,26 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
+  digest = "1:dc62e76fd82a4bd369b92e8adc26d69d18c46d194f183179ac06c92fa4360472"
+  name = "github.com/Pallinder/go-randomdata"
```
MatthewDolan: Do you know why this dependency is getting added?

It seems like it shouldn't be necessary for the tracer to work. It would be good to root it out so we don't add an unnecessary dependency to consumers of the package.

Gopkg.toml
jmacd: can't add a constraint on gogo because there is no go source in the root of this repo. 🤷‍♂️

Gopkg.lock
jmacd: Sorry, stray source file in the directory gets its dependencies picked up by dep.

Gopkg.toml
jmacd: `dep` still confuses me. it seems like you need this here once, to get `dep ensure` to succeed the first tine, after which maybe it can be removed.

collector_client_http.go
jmacd: can't add a constraint on gogo because there is no go source in the root of this repo. 🤷‍♂️

jmacd:
Let's assume we're eliminating gRPC.
I suppose a user might want to use gogo vs protobuf based on other dependencies of their application.
I know that @iredelmeier has wanted to do this too, but I'm afraid we shouldn't block on this work.
@MatthewDolan ^^^

Gopkg.lock
jmacd: Fixed.

Gopkg.toml
jmacd: Removed.

collector_client_http.go
MatthewDolan:
```[[constraint]]
  name = "github.com/gogo/protobuf"
  version = "1.2.1"``` does that not work?

jmacd: IT works now. I'm not sure what sort of state I was in, because it wasn't working, because there's no source in the root of that repo. I do not like `dep`.

jmacd: (Fixed)

jmacd:
I take it by "overt" you mean the requirement to select the dependency in an import statement. ✅

PR #209, opened 2019-04-11
Add Access Token metadata and headers to grpc and http reports

cboppert:
@iredelmeier @frenchfrywpepper @JulianGriggs 

This adds the Access Token as a header and metadata for GRPC and HTTP calls for load balancing.

### DIFF
--- a/collector_client_grpc.go
+++ b/collector_client_grpc.go
@@ -119,7 +119,7 @@ func (client *grpcCollectorClient) Report(ctx context.Context, req reportRequest
 	ctx = metadata.NewOutgoingContext(
 		ctx,
 		metadata.Pairs(
-			lightstep.AccessTokenHeaderKey,
+			AccessTokenHeaderKey,
 			client.accessToken,
 		),
 	)
--- a/collector_client_grpc.go
+++ b/collector_client_grpc.go
@@ -8,6 +8,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	grpcmeta "google.golang.org/grpc/metadata"
 
 	// N.B.(jmacd): Do not use google.golang.org/glog in this package.
 	"github.com/lightstep/lightstep-tracer-common/golang/gogo/collectorpb"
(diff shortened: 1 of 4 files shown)


collector_client_grpc.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	grpcmeta "google.golang.org/grpc/metadata"
```
JulianGriggs: nit: You can remove the alias here and just directly access it via `metadata.NewOutgoingContext(...)`

collector_client_grpc.go
```diff
@@ -114,6 +115,8 @@ func (client *grpcCollectorClient) Report(ctx context.Context, req reportRequest
 	if req.protoRequest == nil {
 		return nil, fmt.Errorf("protoRequest cannot be null")
 	}
+
+	ctx = grpcmeta.NewOutgoingContext(ctx, grpcmeta.Pairs("LightStep-Access-Token", client.accessToken))
```
JulianGriggs: I think this string value should be defined as a constant. I don't have strong feelings about whether there should be one per file or if we should put the header in a shared place like `collector_client.go`

JulianGriggs: Actually, I think we should use `CanonicalHeaderKey()` like the values in collector_client_http.go

cboppert: makes sense

options.go
```diff
@@ -35,11 +35,12 @@ const (
...
-	ComponentNameKey  = "lightstep.component_name"
-	GUIDKey           = "lightstep.guid" // <- runtime guid, not span guid
-	HostnameKey       = "lightstep.hostname"
-	CommandLineKey    = "lightstep.command_line"
+	ParentSpanGUIDKey    = "parent_span_guid" // ParentSpanGUIDKey is the tag key used to record the relationship between child and parent spans.
+	ComponentNameKey     = "lightstep.component_name"
+	GUIDKey              = "lightstep.guid" // <- runtime guid, not span guid
+	HostnameKey          = "lightstep.hostname"
+	CommandLineKey       = "lightstep.command_line"
+	AccessTokenHeaderKey = "LightStep-Access-Token"
```
JulianGriggs: I don't think `options.go` is the right place for this constant. This file is about options on the tracer but this header isn't something that we are making an option. I think `collector_client.go` is a better location. Also, maybe we can use something like https://github.com/lightstep/lightstep-tracer-go/blob/master/collector_client_http.go#L19

Also, let's standardize on using standard header values so `Lightstep-Access-Token` with the `s` lowercased.

PR #213, opened 2019-05-03
allow lazy loggers to emit 0 -> N log entries instead of just 1

MatthewDolan:
R: @bg451 @iredelmeier

CC: @jmacd 

### Summary of Change
I wanted to extend the functionality of the lazy logger to allow emitting 0 fields (for optional fields) or more than one field (for multiple lazy fields computed from the same object).

The one downside I see is that this makes it so the number of fields encoded is not equal to the number of fields passed in. This means that the array can't be pre-allocated to the exact right capacity. I think that's a reasonable trade-off though.

For the expected use case (emitting exactly one field form a lazy logger), this should work mostly as expected. For unexpected use cases, this might be a change in behavior. Before, it was last write wins, but now it will emit all fields.

### DIFF
--- a/proto_logencoder.go
+++ b/proto_logencoder.go
@@ -14,9 +14,9 @@ const (
 
 // An implementation of the log.Encoder interface
 type grpcLogFieldEncoder struct {
-	converter       *protoConverter
-	buffer          *reportBuffer
-	currentKeyValue *collectorpb.KeyValue
+	converter *protoConverter
+	buffer    *reportBuffer
+	keyValues []*collectorpb.KeyValue
 }
 
 func marshalFields(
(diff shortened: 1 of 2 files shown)


proto_logencoder.go
```diff
@@ -28,34 +28,47 @@ func marshalFields(
 	logFieldEncoder := grpcLogFieldEncoder{
 		converter: converter,
 		buffer:    buffer,
+		keyValues: make([]*collectorpb.KeyValue, 0, len(fields)),
 	}
-	protoLog.Fields = make([]*collectorpb.KeyValue, len(fields))
-	for i, field := range fields {
```
bg451: what an unintuitive little piece of code

MatthewDolan:
thanks @bg451!

PR #215, opened 2019-06-17
inject collectorpb.CollectorServiceClient

thepwagner:
This adds the ability to inject a custom protobuf transport.
This requires migrating protobuf conversion from `collectorClient`s to the `tracerImpl`, which is a hard step away from non-Protobuf transports.

I'm running tracers in a network without access to the collector, so I'd like to bring my own transport API. The `grpcCollectorClient` is close, but crufty.

### DIFF
--- a/collector_client.go
+++ b/collector_client.go
@@ -25,31 +25,29 @@ type collectorResponse interface {
 	DevMode() bool
 }
 
-type reportRequest struct {
-	protoRequest *collectorpb.ReportRequest
-	httpRequest  *http.Request
+// Collector encapsulates custom transport of protobuf messages
+type Collector interface {
+	Report(context.Context, *collectorpb.ReportRequest) (*collectorpb.ReportResponse, error)
 }
 
 // collectorClient encapsulates internal grpc/http transports.
 type collectorClient interface {
-	Report(context.Context, reportRequest) (collectorResponse, error)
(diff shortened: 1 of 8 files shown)


austinlparker:
I thought we had actually removed everything but protobuf in the most recent release, but yeah, we're generally in favor of moving all of the tracers towards only using proto.

austinlparker:
The changes look fine overall and I'm fine with the refactor if you want to take a stab at it, or we can merge as-is.

thepwagner:
@austinlparker thanks for the feedback!

I've migrated `protoConverter`, `reporterID` and `attributes` from each `collectorClient` implementation to the `tracerImpl`.
That leaves `collectorClient.Translate()` operating on an already converted protobuf request, but otherwise unchanged. (I toyed with removing `reportRequest` intermediary, but didn't want to mess with the semantics/error behaviour of `Translate`)

Some tangential changes snuck into the HTTP collector too:
- the `context.Context` in `Translate()` is just overridden by `Report()`, so drop it.
- the `accessToken` header can probably be set with the other headers(?)

thepwagner:
Is there any timeline for merging/releasing this?

austinlparker:
Whoops, slipped off my plate. I'll get to merging this today!

austinlparker:
This is released in v0.17.0

PR #221, opened 2019-09-23
Expose record span in the Tracer interface

drolando:
I want to be able to emit a custom made span. `RecordSpan` would allow me to do that easily, however it's not part of the Tracer interface so I can't call it from my code.

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -26,6 +26,8 @@ type Tracer interface {
 	Options() Options
 	// Disable prevents the tracer from recording spans or flushing
 	Disable()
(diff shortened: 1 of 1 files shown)


iredelmeier:
Thanks for the PR!

Do you mind providing more context on what you're aiming to achieve with the custom span? It may be achievable through either existing or new `StartSpanOption`s and/or `FinishOptions`.

In particular, I want to make sure that your use case is supported in OpenTelemetry!

chadbrad:
Doing some housekeeping, closing this, as it's a bit dated. @drolando Please feel free to reach out with more information.

PR #223, opened 2019-09-23
Extend CI no output timeout

iredelmeier:
This prevents spurious failures caused by CircleCI cutting off the build
when it goes too long without output, which happens sometimes since the
tests don't output anything until they've been completed.


iredelmeier:
Closing because this won't go green unless #222 gets merged, and #222 already contains the relevant commit.

PR #227, opened 2019-10-24
Adding B3 header support

codeboten:
This change provides the ability to configure B3 headers propagation via a Propagator option parameter when creating a new Tracer:

```go
lightstepTracer := lightstep.NewTracer(lightstep.Options{
  AccessToken: "YourAccessToken",
  Propagator: "b3",
})
```

The default configuration uses LightStep headers as before. As part of this change, adding a [Propagator interface](https://github.com/lightstep/lightstep-tracer-go/blob/codeboten/b3-headers/propagation.go) to support future formats, currently implemented by [lightstepPropagator](https://github.com/lightstep/lightstep-tracer-go/blob/codeboten/b3-headers/propagation_lightstep.go) and [b3Propagator](https://github.com/lightstep/lightstep-tracer-go/blob/codeboten/b3-headers/propagation_b3.go).

### DIFF
--- a/options.go
+++ b/options.go
@@ -170,6 +170,9 @@ type Options struct {
 	UseHttp bool `yaml:"use_http"`
 	UseGRPC bool `yaml:"usegrpc"`
 
+	// Propagator provides support for different types of headers. Supported options: ls, b3
+	Propagator string `yaml:"propagator"`
+
 	// CustomCollector allows customizing the Protobuf transport.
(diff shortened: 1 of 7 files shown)


chadbrad:
wo0t!

codeboten:
Correct. This is only adding support for b3 header propagation. Implementing a PropagatorStack as in the Java implementation will be done separately.

Issue #226, opened 2019-10-24
Allow `NewTracer` to return error.


bwplotka:
Hi :wave: 

Any particular reason why there is no proper error handling for `NewTracer` constructor? 

I would vote for extending return arg and add `err` (: 

https://github.com/lightstep/lightstep-tracer-go/blob/HEAD/tracer.go#L88

jmacd:
This will require a major version bump.  We could rename the existing method `MustNewTracer()` and panic instead of error. I'd prefer to just add a new method like `NewTracerChecked()`, given that we're putting a lot of effort into the OTel libraries, at this point.

bwplotka:
Oh yea, it definitely can be a separate constructor. Happy to contribute.

jmacd:
We would be glad to accept a pull request for this. Thanks!

PR #228, opened 2019-10-24
Added new constructor for tracer that propagates error.

bwplotka:
Fixes https://github.com/lightstep/lightstep-tracer-go/issues/226

I also made error handling consistent for `NewTracer` - now it emits event only. 

Overall I would suggest totally removing the emit event logic. It is not the best idea to have a global logger, especially for the library: it might cause issues if someone wants some flexibility (e.g running two tracers with different loggers.). Anyway, not a blocker for us here (: 

Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -90,11 +90,21 @@ type tracerImpl struct {
 }
 
 // NewTracer creates and starts a new Lightstep Tracer.
+// In case of error, we emit event and return nil.
 func NewTracer(opts Options) Tracer {
-	err := opts.Initialize()
-	if err != nil {
+	tr, err := CreateTracer(opts)
+	if err != nil  {
 		emitEvent(newEventStartError(err))
 		return nil
+	}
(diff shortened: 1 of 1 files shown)


bwplotka:
Any ideas what's wrong with this test?

iredelmeier:
Hi @bwplotka! The test problem seems to be unrelated to your change.

@codeboten just added #229, which should mitigate the timeout issue.

codeboten:
Looks like we're occasionally bumping up against the default `go test` 10 minute timeout. Opened a PR (https://github.com/lightstep/lightstep-tracer-go/pull/229) to address by bumping the timeout for now. Mind rebasing and giving your branch another try once that PR is merged?

antonio:
@iredelmeier @codeboten is there a timeline for a new release that includes this patch?

PR #230, opened 2019-10-25
Change OC -> LS ID conversion logic

iredelmeier:
This fixes a problem with the opencensus ID conversion logic, since `binary.Uvarint` wasn't capturing the value as expected.

### DIFF
--- a/lightstepoc/internal/conversions/package.go
+++ b/lightstepoc/internal/conversions/package.go
@@ -7,31 +7,19 @@ import (
 	"go.opencensus.io/trace"
 )
 
-func ConvertTraceID(original trace.TraceID) (uint64, bool) {
-	if traceID, n := binary.Uvarint(original[8:]); traceID != 0 && n > 0 {
-		return traceID, true
-	}
-	return 0, false
+func ConvertTraceID(original trace.TraceID) uint64 {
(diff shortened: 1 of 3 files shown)


lightstepoc/internal/conversions/package.go
```diff
@@ -7,31 +7,19 @@ import (
 	"go.opencensus.io/trace"
 )
 
-func ConvertTraceID(original trace.TraceID) (uint64, bool) {
-	if traceID, n := binary.Uvarint(original[8:]); traceID != 0 && n > 0 {
-		return traceID, true
-	}
-	return 0, false
+func ConvertTraceID(original trace.TraceID) uint64 {
+	return binary.BigEndian.Uint64(original[:])
```
jmacd: This should use `original[:8]` I think.  (Are we sure the other OC->LS exporters are behaving exactly the same?)

iredelmeier: I believe this is at least the behaviour in Java, per [this line](https://github.com/lightstep/lightstep-census-java/blob/540c182e1651bfd336085292a717b82add75d89d/src/main/java/com/lightstep/opencensus/exporter/LightStepExporterHandler.java#L168).

PR #232, opened 2019-10-28
Pad least significant bits of trace ID

codeboten:
Moved padding from left to right. Also ensuring the TraceID is not modified as it gets propagated if it started out as 128-bit

### DIFF
--- a/raw_span.go
+++ b/raw_span.go
@@ -38,6 +38,9 @@ type SpanContext struct {
 	// A probabilistically unique identifier for a [multi-span] trace.
 	TraceID uint64
 
+	// Least significant bits of TraceID
+	TraceIDLower uint64
+
 	// A probabilistically unique identifier for a span.
 	SpanID uint64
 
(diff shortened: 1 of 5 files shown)


tracer_transport_test.go
```diff
@@ -592,7 +593,7 @@ var _ = Describe("Tracer Transports", func() {
 				Expect(context).To(BeEquivalentTo(knownContext3))
 			})
 
-			It("should drop higher 64 bits of a 128 bit TraceID", func() {
+			It("should error with an invalid TraceID", func() {
```
iredelmeier: Nit: Might be adding a couple words in the description about what makes this ID invalid. At a glance, it looks like it should be fine, and I think is actually the same ID used in a "valid" test above.

raw_span.go
```diff
@@ -38,6 +38,9 @@ type SpanContext struct {
 	// A probabilistically unique identifier for a [multi-span] trace.
 	TraceID uint64
 
+	// Least significant bits of TraceID
+	TraceIDLower uint64
```
iredelmeier: Do we need to add the field yet, or can we just drop the lower bits for now? Doesn't look like we have anything consuming the new `TraceIDLower`, e.g., for propagation.

tracer_transport_test.go
codeboten: I'll update it, the ID is 31 characters instead of 32 as the one above.

raw_span.go
codeboten: The field gets propagated is used in the b3 propagation: https://github.com/lightstep/lightstep-tracer-go/pull/232/files#diff-c7e8819bc83013a318340945ad302ef3R36

iredelmeier: Doh!

Issue #224, opened 2019-10-09
Add B3 header support 


chadbrad:
Please add B3 header support :) 

Please see this doc for the problem statement and specifics: https://docs.google.com/document/d/1ukvnSoXkavcojlVXpCRzep8PS_guXly_NbbhHIDgbFY/edit

This is high priority task, we need this ASAP. Please note the spec also covers trace-context and envoy support, but those are lower priority and captured in separate git issues. 

CC @tedsuo

ledor473:
@codeboten is there any particular reason why `Propagator` in `options.go` is a String and directly the [underlying structure you're converting the String to](https://github.com/lightstep/lightstep-tracer-go/pull/227/files#diff-46db78603208f38452c0333bec12ab2cR143-R153)?

It makes it pretty hard to extend the capability of the Go tracer compared to the [Java](https://github.com/lightstep/lightstep-tracer-java-common/issues/56) and [NodeJS](https://github.com/lightstep/lightstep-tracer-javascript/commit/d46e32a97549decfbadf22b82e030ce4118d22b2#diff-04c6e90faac2675aa89e2176d2eec7d8R106) implementation.

The current implementation also doesn't allow to have multiple propagators (B3 + LS) at the same time to ease the migration from one to the other.

codeboten:
@ledor473 thanks for the quick feedback! This was an oversight on my part and I created issue #238 to track the work to update this. I'll see about getting a PR reviewed today.

ledor473:
That's awesome! Thanks @codeboten 👍

codeboten:
Here's the PR https://github.com/lightstep/lightstep-tracer-go/pull/239

PR #236, opened 2019-11-05
Don't pad since b3 supports 64-bit as well as 128-bit trace IDs

codeboten:
This change removes the padding for 64-bit trace IDs as it is no longer needed. In the event of a 128-bit ID that needs to be truncated to 64-bit, the most significant bits will be truncated.

Signed-off-by: Alex Boten <aboten@lightstep.com>


codeboten:
Closes #224

PR #239, opened 2019-11-14
Use map instead of string to configure propagators

codeboten:
This enables users to configure custom propagators.

Signed-off-by: Alex Boten <aboten@lightstep.com>

### DIFF
--- a/propagation_b3.go
+++ b/propagation_b3.go
@@ -14,7 +14,8 @@ const (
 	b3FieldNameSampled = b3Prefix + "sampled"
 )
 
-var theB3Propagator b3Propagator
+// B3Propagator propagates context in the b3 format
+var B3Propagator b3Propagator
 
 type b3Propagator struct{}
 
--- a/tracer.go
+++ b/tracer.go
@@ -93,7 +93,7 @@ type tracerImpl struct {
 // In case of error, we emit event and return nil.
 func NewTracer(opts Options) Tracer {
 	tr, err := CreateTracer(opts)
-	if err != nil  {
+	if err != nil {
 		emitEvent(newEventStartError(err))
 		return nil
 	}
@@ -154,10 +154,10 @@ func CreateTracer(opts Options) (Tracer, error) {
 		opentracing.HTTPHeaders: theLightStepPropagator,
 		opentracing.Binary:      theBinaryPropagator,
 	}
-
-	if opts.Propagator == "b3" {
-		impl.propagators[opentracing.TextMap] = theB3Propagator
-		impl.propagators[opentracing.HTTPHeaders] = theB3Propagator
+	if len(opts.Propagators) > 0 {
+		for builtin, propagator := range opts.Propagators {
+			impl.propagators[builtin] = propagator
+		}
 	}
 
 	return impl, nil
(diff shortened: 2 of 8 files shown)


codeboten:
closes #239

tracer.go
```diff
@@ -154,10 +154,10 @@ func CreateTracer(opts Options) (Tracer, error) {
 		opentracing.HTTPHeaders: theLightStepPropagator,
 		opentracing.Binary:      theBinaryPropagator,
 	}
-
-	if opts.Propagator == "b3" {
-		impl.propagators[opentracing.TextMap] = theB3Propagator
-		impl.propagators[opentracing.HTTPHeaders] = theB3Propagator
+	if len(opts.Propagators) > 0 {
+		for builtin, propagator := range opts.Propagators {
```
bg451: For what its worth you don't need the if statement, you can range over a nil map or empty map just fine.

codeboten: good point, fixed

codeboten:
@iredelmeier I think you're right, the plan was to eventually implement a PropagatorStack to support multiple propagators as the java implementation: https://github.com/lightstep/lightstep-tracer-java-common/blob/master/common/src/main/java/com/lightstep/tracer/shared/PropagatorStack.java

codeboten:
Though re-reading the implementation in [charp](https://github.com/lightstep/lightstep-tracer-csharp/blob/b1170523cf838ba3fc6c99e4e4f7cacd6923224a/src/LightStep/Propagation/PropagatorStack.cs#L9) and [java](https://github.com/lightstep/lightstep-tracer-java-common/blob/2fdfe4acc382a0a2d66f6c9dcfe3fc4579420bb3/common/src/main/java/com/lightstep/tracer/shared/PropagatorStack.java#L21), the interface for PropagatorStack is the same as a Propagator, so this change works.

propagation_b3.go
```diff
@@ -14,7 +14,8 @@ const (
 	b3FieldNameSampled = b3Prefix + "sampled"
 )
 
-var theB3Propagator b3Propagator
+// B3Propagator propagates context in the b3 format
+var B3Propagator b3Propagator
```
ledor473: Should `theLightStepPropagator` be exported as well?

iredelmeier:
@codeboten I think my understanding is the same as yours re: Java and C#? That is, a PropagatorStack is still a Propagator, but it's implemented as an array of n propagators.

The split across different propagator formats makes this a bit more painful... :/

codeboten:
Right, the difference being with the PropagatorStack, the format has to be specified at creation, where as with this implementation, which is similar to the javascript/ruby implementations, a map of formats can be specified and propagators for each format configured.

I suggest merging this change, and in a later change providing an implementation of the PropagatorStack to support multiple propagators per format.

codeboten:
Added a PropagatorStack to support multiple formats.

codeboten:
Good question, we could support something like
```
	 propagators:
	 	http_headers:
	 		ls
	 		b3
	 	text_map:
	 		ls
	 	binary:
	 		binary
```
Although that wouldnt account for custom propagators

PR #241, opened 2019-11-26
Add the option to add default tags.

bviolier:
Adds the ability to set default tags when creating the exporter.

### DIFF
--- a/lightstepoc/options.go
+++ b/lightstepoc/options.go
@@ -14,6 +14,9 @@ const (
 // Option provides configuration for the Exporter
 type Option func(*config)
 
+// Attributes provide extra information in a span
+type Attributes map[string]interface{}
+
 // WithAccessToken sets an access token for communicating with LightStep
 func WithAccessToken(accessToken string) Option {
 	return func(c *config) {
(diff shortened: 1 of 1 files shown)


lightstepoc/options.go
```diff
@@ -60,6 +60,15 @@ func WithComponentName(componentName string) Option {
 	}
 }
 
+// WithDefaultTags sets tags that will be appended to every span that is exported to the trace.
```
iredelmeier: I think we should rename this to `WithDefaultAttributes` to be consistent with OpenCensus's naming. ("Tag" means something else to them; "attribute" is more akin to OpenTracing's tags.)

lightstepoc/options.go
```diff
@@ -60,6 +60,15 @@ func WithComponentName(componentName string) Option {
...
+func WithDefaultTags(tags opentracing.Tags) Option {
+	return func(c *config) {
+		for key, value := range tags {
```
iredelmeier: I think we should rename this to WithDefaultAttributes to be consistent with OpenCensus's naming. ("Tag" means something else to them; "attribute" is more akin to OpenTracing's tags.)

iredelmeier:
Thanks @bviolier! This looks great to me aside from the suggested naming change.

"Attributes" aren't really clarified in the OC spec, so here's the `SpanData` type: https://github.com/census-instrumentation/opencensus-go/blob/aad2c527c5defcf89b5afab7f37274304195a6b2/trace/export.go#L73-L97

bviolier:
Thanks, that makes sense! I tried to actually use trace.Attribute now, but the values are private, so I can't read them back to put them in Tags, unfortunately. See updated PR, let me know if this is ok!

PR #242, opened 2019-12-06
Add flush duration to status report struct

stephen:
The flush duration is the amount of time it took for the tracer to pack up the records, send them over the network, and rotate buffers. Seeing this measurement is helpful for tuning the min reporting period for the tracer.

Because the reporter loop is serial, this is the minimum time between flushes that the tracer could do. Add a test that this value is measured.

### DIFF
--- a/events.go
+++ b/events.go
@@ -147,12 +147,29 @@ func (e *eventConnectionError) Err() error {
 type EventStatusReport interface {
 	Event
 	EventStatusReport()
+
+	// StartTime is the earliest time a span was added to the report buffer.
 	StartTime() time.Time
+
+	// FinishTime is the latest time a span was added to the report buffer.
 	FinishTime() time.Time
+
+	// Duration is time between StartTime and FinishTime.
 	Duration() time.Duration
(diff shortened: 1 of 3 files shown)


andrewcheelightstep:
@bg451 Can you review this?

events.go
```diff
@@ -196,6 +200,10 @@ func (s *eventStatusReport) Duration() time.Duration {
 	return s.finishTime.Sub(s.startTime)
 }
 
+func (s *eventStatusReport) FlushDuration() time.Duration {
```
bg451: Could you add a comment that mentions the flush duration is time from flush start to rotating out the buffers, not necessarily just the latency of the network call to flush spans?

stephen: sg, though i think it won't be too helpful to end users unless we add a comment to the interface too - wdyt?

stephen: @bg451, I added godoc comments for each field on the interface, wdyt?

bg451: yep! This is great, thanks!

PR #158, opened 2018-09-06
there is no OnEvent static method anymore 

etsangsplk:
OnEvent is of this signature, rather than in the original README
https://github.com/lightstep/lightstep-tracer-go/blob/64af8e62274eb002d27a0e1ddb74f936cbbe3d8f/event_handlers.go#L69

Issue #157, opened 2018-09-05
There is no OnEvent static method ad in README


etsangsplk:
Looks like you guys switched to using 
```
lightstep.SetGlobalEventHandler(logTracerEventHandler)
```

https://github.com/lightstep/lightstep-tracer-go/blob/644c3d5ecbd0499c50a1329f89ba287921fc1144/README.md

iredelmeier:
Fixed with #158

PR #243, opened 2019-12-08
update README.md to replace mention of OnEvent

andrewhsu:
The `OnEvent` function is deprecated, use `SetGlobalEventHandler` instead.

Follow up to PR #158
Related issue #157


iredelmeier:
Thanks @andrewhsu!

PR #249, opened 2020-03-10
Adding metrics

codeboten:
Still some work needed:

- needs retry/backoff logic
- more unit tests
- add duration


tedpennings:
Metric names are 💯  - thanks !

events.go
```diff
@@ -353,3 +353,58 @@ func (eventTracerDisabled) EventTracerDisabled() {}
...
+func newEventSystemMetricsMeasurementFailed(err error) *eventSystemMetricsMeasurementFailed {
+	return &eventSystemMetricsMeasurementFailed{
+		err: err,
+	}
+}
+
+func (e *eventSystemMetricsMeasurementFailed) Event() {}
+
+func (e *eventSystemMetricsMeasurementFailed) String() string {
+	return e.err.Error()
```
iredelmeier: nit: Either `#String()` or `#Error()` should probably piggyback off the other

Gopkg.lock
```diff
@@ -1,258 +0,0 @@
-# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
iredelmeier: Suggestion: extract the Gopkg.lock/Gopkg.toml removal into a separate PR

examples/trivial.go
```diff
@@ -7,17 +7,17 @@ package main
...
+	"os"
 	"time"
 
 	"github.com/lightstep/lightstep-tracer-go"
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/log"
 )
 
-var accessToken = flag.String("access_token", "", "your LightStep access token")
+var accessToken = os.Getenv("LIGHTSTEP_ACCESS_TOKEN")
```
iredelmeier: Suggestion: as with Gopkg.lock/Gopkg.toml, move these changes to a separate PR

go.sum
```diff
@@ -39,15 +35,22 @@ github.com/hashicorp/golang-lru v0.5.1 h1:0hERBMJE1eitiLkihrMvRVBYAkpHzc/J3QdDN+
 github.com/hashicorp/golang-lru v0.5.1/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/iredelmeier/lightstep-tracer-common/golang/gogo v0.0.0-20200214020106-93e2f3cec85e h1:YiQ+VsjS97Q935VOAOyOw76V7RN+QO5V2aCKRrv0ihc=
```
iredelmeier: Not sure why references to my fork are still in here. Maybe try `go mod tidy`?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+type ProcessCPU struct {
+	User   float64
+	System float64
+}
+
+type CPU struct {
+	User   float64
+	System float64
+	Usage  float64
+	Total  float64
```
iredelmeier: I think originally there was also interest in capturing nice and/or steal. Is this no longer needed?

(Also can  definitely be added later!)

internal/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+type Metrics struct {
+	ProcessCPU       ProcessCPU
+	CPU              map[string]CPU
+	NIC              map[string]NIC
+	Memory           Memory
+	CPUPercent       float64
+	GarbageCollector GarbageCollector
+}
+
+type GarbageCollector struct {
```
iredelmeier: (Can be addressed later) Guessing we also want other GC-related metrics here, e.g., `NextGC` and `LastGC`.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+		usage := t.User + t.System + t.Nice + t.Iowait + t.Irq + t.Softirq + t.Steal
+		metrics.CPU[t.CPU] = CPU{
+			User:   t.User,
+			System: t.System,
+			Usage:  usage,
+			Total:  usage + t.Idle,
+		}
+	}
+
+	for _, p := range percentages {
```
iredelmeier: Might be clearer to assert that `len(percentages) == 1`, and then only use `percentages[0]`.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return Metrics{}, err
+	}
+
+	metrics := Metrics{
+		ProcessCPU: ProcessCPU{
+			User:   processTimes.User,
+			System: processTimes.System,
+		},
+		CPU: make(map[string]CPU),
```
iredelmeier: Small suggestion: since we already known the sizes, we can use:
```suggestion
		CPU: make(map[string]CPU, len(systemTimes)),
		NIC: make(map[string]NIC, len(netStats)),
```

internal/metrics/reporter.go
```diff
@@ -0,0 +1,326 @@
...
+	for k, v := range attributes {
+		tags = append(tags, &collectorpb.KeyValue{Key: k, Value: &collectorpb.KeyValue_StringValue{StringValue: v}})
+	}
+	return tags
+}
+
+func getLabels(attributes map[string]string) []*collectorpb.KeyValue {
+	labels := []*collectorpb.KeyValue{}
+	filters := []string{
+		"lightstep.component_name",
```
iredelmeier: nit: use the constants from `options.go` (or is there a circular dependency? ...argh)

internal/metrics/reporter.go
```diff
@@ -0,0 +1,326 @@
...
+		pb.Points = append(pb.Points, r.addFloat("net.bytes_recv", float64(nic.BytesReceived-r.stored.NIC[label].BytesReceived), metricspb.MetricKind_COUNTER, intervals))
+		pb.Points = append(pb.Points, r.addFloat("net.bytes_sent", float64(nic.BytesSent-r.stored.NIC[label].BytesSent), metricspb.MetricKind_COUNTER, intervals))
+	}
+
+	err = r.send(ctx, pb)
+	if err != nil {
+		return err
+	}
+
+	r.stored = m
```
iredelmeier: Suggestion: access these mutable fields behind a `sync.RWMutex`. My brain's pretty fuzzy right now, though; I'm trying to figure out what will happen if, e.g., a previous request is still getting retried when a new one comes in (especially if the timeout and/or metric duration timeouts have been overridden).

internal/metrics/reporter_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	JustBeforeEach(func() {
+		h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			body, _ := ioutil.ReadAll(r.Body)
+			err := proto.Unmarshal(body, &ingestRequest)
+			if !Expect(err).To(BeNil()) {
+				return
+			}
+		})
+		s := httptest.NewServer(h)
```
iredelmeier: Does this server not get closed? Looks like it leaks across tests?

internal/metrics/reporter_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/lightstep/lightstep-tracer-go/internal/metrics"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Reporter", func() {
+	var reporter *metrics.Reporter
+	var ingestRequest metricspb.IngestRequest
+
+	JustBeforeEach(func() {
```
iredelmeier: General suggestion: avoid `JustBeforeEach` if you can help it. IME, it always leads to more pain than it's worth - confusing control flows, weird race conditions, enabling of overly complicated tests...

internal/metrics/reporter_test.go
```diff
@@ -0,0 +1,110 @@
...
+			for _, point := range points {
+				name := point.GetMetricName()
+				Expect(point.Kind).To(Equal(expected[name]))
+			}
+		})
+	})
+	Describe("Measure fails unretryably", func() {
+		It("should return an error", func() {
+			h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusNotFound)
```
iredelmeier: Suggestion: table tests for all non-retryable error codes

internal/metrics/reporter_test.go
```diff
@@ -0,0 +1,110 @@
...
+			reporter.Measure(context.Background(), 1)
+			err := reporter.Measure(context.Background(), 1)
+			Expect(err).To(Not(BeNil()))
+			Expect(err.Error()).To(ContainSubstring("404"))
+		})
+	})
+	Describe("Measure fails retryably", func() {
+		It("should return after retrying", func() {
+			h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusBadGateway)
```
iredelmeier: Suggestion: table tests for all retryable error codes

internal/metrics/reporter.go
```diff
@@ -0,0 +1,326 @@
...
+		return err
+	}
+
+	req = req.WithContext(ctx)
+
+	req.Header.Set(contentTypeHeader, protoContentType)
+	req.Header.Set(acceptHeader, protoContentType)
+	req.Header.Set(accessTokenHeader, r.accessToken)
+
+	if !r.skippedInitialReport {
```
iredelmeier: Does this mean that no metrics will get sent if the tracer is closed after either <30s or <1m?

go.sum
codeboten: Fixed

internal/metrics/metrics.go
codeboten: Correct, these are now being summed as part of usage rather than individually:
https://github.com/lightstep/lightstep-tracer-go/pull/249/files/a847ca7d4d70b7b8b5a798d42c6bc6e3554d22f5#diff-501d4915da42d43fcfa855ffcc089cfeR104

internal/metrics/metrics.go
codeboten: Thanks for reminding me, percentages have been removed entirely, updated the code.

internal/metrics/reporter_test.go
codeboten: Good catch. Fixed

internal/metrics/reporter_test.go
codeboten: I'm not too familiar with ginko, does `BeforeEach` also suffer from this?

Gopkg.lock
codeboten: Moved into a separate PR, will merge it into this branch once its in master

events.go
codeboten: Fixed. Updated the rest of the events as well

internal/metrics/reporter.go
codeboten: it was a circular dep, i added a `constants` package to move out at least the common keys

constants/constants.go
```diff
@@ -0,0 +1,10 @@
+package constants
+
+const (
+	// ComponentNameKey is the tag key identifying the service
+	ComponentNameKey = "lightstep.component_name"
```
karkum: can we re-use https://github.com/lightstep/lightstep-tracer-go/blob/master/options.go#L39?

constants/constants.go
```diff
@@ -0,0 +1,10 @@
...
+	// HostnameKey is the tag key identifying hostname
+	HostnameKey = "lightstep.hostname"
```
karkum: can we re-use https://github.com/lightstep/lightstep-tracer-go/blob/master/options.go#L41?

constants/constants.go
codeboten: i created the constants package to allow both metrics and the lightstep package to share constants. otherwise it was creating an import cycle when trying to import options from the metrics package

internal/metrics/reporter_test.go
iredelmeier: Nope :) `BeforeEach` behaves more consistently with the equivalents from other test harnesses out there. Basically, both run FIFO, but `JustBeforeEach`s run _after_ all `BeforeEach`s... which means that an outer layer's `JustBeforeEach` can override an inner layer's `BeforeEach`.

```go
var _ = Describe("grandparent", func() {
	BeforeEach(func() {
		fmt.Println("grandparent BeforeEach")
	})

	JustBeforeEach(func() {
		fmt.Println("grandparent JustBeforeEach")
	})

	Describe("parent", func() {
		BeforeEach(func() {
			fmt.Println("parent BeforeEach")
		})

		JustBeforeEach(func() {
			fmt.Println("parent JustBeforeEach")
		})

		Describe("child", func() {
			BeforeEach(func() {
				fmt.Println("child BeforeEach")
			})

			JustBeforeEach(func() {
				fmt.Println("child JustBeforeEach")
			})

			It("runs", func() {
				Expect(1 + 1).To(Equal(2))
			})
		})
	})
})
```

prints:
```
grandparent BeforeEach
parent BeforeEach
child BeforeEach
grandparent JustBeforeEach
parent JustBeforeEach
child JustBeforeEach
```

internal/metrics/reporter_test.go
```diff
@@ -19,21 +19,50 @@ import (
...
 	var ingestRequest metricspb.IngestRequest
+	var okServer *httptest.Server
+	var okURL string
+	var retryableServer *httptest.Server
+	var retryableURL string
+	var unretryableServer *httptest.Server
+	var unretryableURL string
 
-	JustBeforeEach(func() {
+	BeforeSuite(func() {
```
iredelmeier: `BeforeSuite` is another weird one. A "suite" is a whole package's tests, rather than a `Describe` block's tests. Consequently, you can only have one `BeforeSuite` per package, and it's almost always at the topmost level.

constants/constants.go
iredelmeier: Thoughts on sticking these into an `internal` package? (Moving them out of the top package would already constitute a breaking change.)

options.go
```diff
@@ -266,6 +282,26 @@ func (opts *Options) Initialize() error {
 		}
 	}
 
+	if opts.SystemMetrics.Endpoint.Host == "" {
+		opts.SystemMetrics.Endpoint.Host = DefaultSystemMetricsHost
```
karkum: @codeboten did we consider making this default to the same endpoint as the collector? That would simplify the config even more for our internal services

internal/metrics/reporter.go
```diff
@@ -0,0 +1,338 @@
...
+
+	pb, err := r.prepareRequest(m)
+	if err != nil {
+		return err
+	}
+
+	pb.Points = append(pb.Points, r.addFloat("runtime.go.cpu.user", m.ProcessCPU.User-r.stored.ProcessCPU.User, metricspb.MetricKind_COUNTER, intervals))
+	pb.Points = append(pb.Points, r.addFloat("runtime.go.cpu.sys", m.ProcessCPU.System-r.stored.ProcessCPU.System, metricspb.MetricKind_COUNTER, intervals))
+	pb.Points = append(pb.Points, r.addFloat("runtime.go.gc.count", float64(m.Runtime.NumGC-r.stored.Runtime.NumGC), metricspb.MetricKind_COUNTER, intervals))
+	pb.Points = append(pb.Points, r.addFloat("runtime.go.goroutine", float64(m.Runtime.NumGoroutine-r.stored.Runtime.NumGoroutine), metricspb.MetricKind_COUNTER, intervals))
```
karkum: @codeboten i think this should be a gauge - seeing extremely large numbers on the read side

codeboten: Good catch, yes this should be a gauge.

examples/trivial.go
codeboten: Reverted

internal/metrics/reporter.go
codeboten: From what I was seeing, the context will timeout and return an error before another request is created

PR #263, opened 2020-09-30
Propagate sampled flag

### LINKED ISSUE #261: Sampling decision propagation
It doesn't appear that the library supports sampling decision propagation. It doesn't take the sampled flag from the carrier, nor propagate it locally nor used it in collection.

ocelotl:
Fixes #261

### DIFF
--- a/tracer_test.go
+++ b/tracer_test.go
@@ -359,6 +359,18 @@ var _ = Describe("Tracer", func() {
 			})
 		})
 
+		Context("when the span context is sampled", func() {
+			It("the span should be recorded", func() {
+				span := tracer.StartSpan("parent", SetTraceID(1), SetSampled("true"))
+				span.Finish()
+				Expect(fakeRecorder.RecordSpanCallCount()).To(Equal(1))
+
+				span = tracer.StartSpan("parent", SetTraceID(1), SetSampled("fals"))
+				span.Finish()
+				Expect(fakeRecorder.RecordSpanCallCount()).To(Equal(2))
+			})
+		})
+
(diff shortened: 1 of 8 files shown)


kayousterhout:
Do you need to set Sampled here (instead of always using true): https://github.com/lightstep/lightstep-tracer-go/blob/master/propagation_lightstep.go#L40

ocelotl:
> Do you need to set Sampled here (instead of always using true): https://github.com/lightstep/lightstep-tracer-go/blob/master/propagation_lightstep.go#L40

Right, fixed. :+1:

tracer_test.go
```diff
@@ -348,6 +348,31 @@ var _ = Describe("Tracer", func() {
...
+				span.Finish()
+				Expect(fakeRecorder.RecordSpanCallCount()).To(Equal(0))
+			})
+		})
+
+		Context("when the parent span carries the Sampled flag", func() {
+			It("should propagate flags from parent to child", func() {
+				opentracing.SetGlobalTracer(tracer)
+
+				opentracing_parent_span := tracer.StartSpan("parent", SetTraceID(1), SetSampled("false"))
```
codeboten, line 366: Please add a test here to ensure that if a span is started with the sampled set to anything else than false, the child span will be recorded.

codeboten (changes requested):
Change looks good, just one comment around one additional test that I think we should have.

tracer_test.go
ocelotl: Fixed

codeboten (approved):
👍

PR #267, opened 2020-12-09
LS-21346: lightstep-tracer-go: generate uint64 trace IDs, not int64

neena:
R: @codeboten 
CC: @lightstep/team-ingest 

We discovered that the tracer is generating int64 trace IDs and converting them to uint64s, which means we're only using half of the trace ID space! This PR makes it so we actually generates random uint64 trace IDs. (This should halve our probability of trace ID collisions).

### DIFF
--- a/lightstep/rand/locked_rand.go
+++ b/lightstep/rand/locked_rand.go
@@ -60,6 +60,15 @@ func (lr *LockedRand) Uint64() (n uint64) {
 	return
 }
 
+// TwoUint64 generates 2 random uint64 without locking twice.
+func (lr *LockedRand) TwoUint64()  (n1, n2 uint64) {
+	lr.lk.Lock()
+	n1 = lr.r.Uint64()
+	n2 = lr.r.Uint64()
+	lr.lk.Unlock()
+	return
+}
+
 // Int31 returns a non-negative pseudo-random 31-bit integer as an int32.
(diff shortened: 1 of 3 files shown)


lightstep/rand/locked_rand.go
```diff
@@ -60,6 +60,15 @@ func (lr *LockedRand) Uint64() (n uint64) {
 	return
 }
 
+// TwoUint64 generates 2 random uint64 without locking twice.
+func (lr *LockedRand) TwoUint64()  (n1, n2 uint64) {
+	lr.lk.Lock()
+	n1 = lr.r.Uint64()
+	n2 = lr.r.Uint64()
+	lr.lk.Unlock()
```
MatthewDolan, line 68: nit: Both here and above, I this it would be preferred to have the `Unlock()` calls be `defer`-ed.

MatthewDolan (approved):
LGTM.

lightstep/rand/locked_rand.go
neena: I wanted to keep consistency with the other functions in this file!

neena:
@codeboten did you get a chance to look at this? do you have any concerns, or can I go ahead with this?

codeboten (approved):
👍

Issue #273, opened 2021-02-09
latency issue caused by MaxBufferedSpans Option


szuecs:
In our http proxy https://github.com/zalando/skipper we see an increase in Heap object creation that causes GC p75 slightly above 1ms and this seems to cause latency p999 <=25ms to be impossible. See also the blue line in the graph that changed close to 1/28 and is ok again around 2/9. The only option I changed was MaxBufferedSpans option set from 32768 back to 8192. We had 8192 before 1/28. On 1/28 we also changed the tracer library from v0.16.0 to v0.24.0.
We had to increase MaxBufferedSpans setting, because there was 20-30% client side Span drops.

![image](https://user-images.githubusercontent.com/50872/107375386-f9c00500-6ae8-11eb-819f-fbfae22ab9c4.png)
![image](https://user-images.githubusercontent.com/50872/107376075-c29e2380-6ae9-11eb-9c7c-bc0c97591f7d.png)
![image](https://user-images.githubusercontent.com/50872/107376242-f4af8580-6ae9-11eb-8ad2-e03165893074.png)
 

We built skipper with  Go v1.15.7  and lightstep-tracer-go v0.24.0.

I hoped the tracing buffer would be a pre-allocated ring buffer, that would consume only a fixed amount of memory without creating garbage.

szuecs:
I have another occurrence. Interesting is that this has only 400 RPS, which is basically nothing. And instances differ with close to 10x CPU and memory consumption:

CPU usage per instance:
![image](https://user-images.githubusercontent.com/50872/109027016-a6d97680-76c0-11eb-8b76-8753c37820f2.png)

Memory allocations per instance:
![image](https://user-images.githubusercontent.com/50872/109026939-91fce300-76c0-11eb-8982-033bcdfdf0a5.png)

memory profile:
![image](https://user-images.githubusercontent.com/50872/109026751-5cf09080-76c0-11eb-80a5-ab6baf0e571b.png)

Another instance that does not have the same problem, yet:
![image](https://user-images.githubusercontent.com/50872/109027514-28310900-76c1-11eb-9abc-87a250126f39.png)

szuecs:
I copy #276 into this one, because it's the same problem.

### original issue

We run skipper with deployment+hpa as ingress controller in Kubernetes. In front of this there is an ALB distributing the traffic to the skipper. Maybe related to #273.

```
kubectl top pods -l application=skipper-ingress 
[...]

PR #277, opened 2021-03-15
Fix/span drops message larger than max

### LINKED ISSUE #275: UI span drops in stream but not showing up in reporting status page
stream:
image

Reporting Status page says:
- Go v1.15.7
- lightstep tracer v0.24.0
- instances: 417
- Spans sent: 62,338,477
- client dropped: 0-355,785 (normally around 10k)
- satellite dropped: 0-55,240 (normally around 5k)

I see different errors in logs:
```
2021/02/16 14:14:13 LS Tracer error: (rpc error: code = Unavailable desc = upstream connect error or disconnect/reset before headers. reset reason: local reset). NOTE: Set the SetGlobalEventHandler handler to log events.
2021/02/16 14:37:28 LS Tracer error: (rpc error: code = ResourceExhausted desc = trying to send message larger than max (18688051 vs. 16384000)). NOTE: Set the SetGlobalEventHandler handler to log events.
2021/02/16 16:17:34 LS Tracer error: (rpc error: code = ResourceExhausted desc = trying to send message larger than max (17011037 vs. 16384000)). NOTE: Set the SetGlobalEventHandler handler to log events.
2021/02/16 16:22:27 LS Tracer error: (Post "https://ingest.lightstep.com:443/metrics": read tcp 172.31.1.167:42960->35.222.202.121:443: read: connection reset by peer). NOTE: Set the SetGlobalEventHandler handler to log events.
```

Most relevant seems to me the `ResourceExhausted desc` messages.
If one of these `rpc error:` messages appear the tracer is not sending any Spans anymore to satellites and reporting in the lightstep UI won't count the dropped spans at all.

We are using the following configuration:
```go
	lightstep.Options{
...
		UseGRPC:                     true,
		Tags:                        tags,
		MaxBufferedSpans:            32768,
		GRPCMaxCallSendMsgSizeBytes: 16384000,
		ReportingPeriod:             "2500ms"
		MinReportingPeriod:          "500ms",
		Propagators:                 "lightstep", // use lightstep.LightStepPropagator not B3Propagator
 ```

Any suggestion to have a more healthy configuration would be great.
I wonder if the library could fix the issue for us instead of we figuring out the right configuration every now and then (happened to me already the second time in one year).

szuecs:
Fix #275 and likely also #273. 

I tested with big span sizes, that had to be split into 2-5 chunks and 1k rps constant traffic a local modified build of [skipper](https://github.com/zalando/skipper), that used the proposed modification. The algorithm to compute the number of parts is: pick 2 items at random and use the bigger one to compute the heuristic of average span size, that will be multiplied by number of spans.
This algorithm shows good results (at 12:30 the patched skipper with the 2-at-random was deployed, before I tried 1-at-random):

image

### DIFF
--- a/collector_client.go
+++ b/collector_client.go
@@ -38,13 +38,14 @@ type reportRequest struct {
 // SplitByParts splits reportRequest into given number of parts.
 // Beware, that parts=0 panics.
 func (rr reportRequest) SplitByParts(parts int) []reportRequest {
-	spans := rr.protoRequest.Spans
-	if len(spans) == 0 {
+
+	if rr.protoRequest == nil || len(rr.protoRequest.Spans) == 0 || parts <= 1 {
 		return []reportRequest{rr}
 	}
+	spans := rr.protoRequest.Spans
 
 	maxSize := len(rr.protoRequest.Spans) / parts
-	if len(rr.protoRequest.Spans) % parts > 0 {
+	if len(rr.protoRequest.Spans)%parts > 0 {
 		maxSize++
 	}
 
--- a/collector_client.go
+++ b/collector_client.go
@@ -35,6 +35,36 @@ type reportRequest struct {
 	httpRequest  *http.Request
 }
 
+// SplitByParts splits reportRequest into given number of parts.
+// Beware, that parts=0 panics.
+func (rr reportRequest) SplitByParts(parts int) []reportRequest {
+	spans := rr.protoRequest.Spans
+	if len(spans) == 0 {
+		return []reportRequest{rr}
+	}
+
+	maxSize := len(rr.protoRequest.Spans) / parts
+	if len(rr.protoRequest.Spans) % parts > 0 {
+		maxSize++
+	}
+
+	var rrs []reportRequest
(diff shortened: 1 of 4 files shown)


szuecs:
It would be great if you can guide me fixing the tests, because half of them are broken and I think there are assumptions in tests that hold not true anymore. Basically right now Flush() returns fast in case there are no rawSpans and it seems 50% of the tests rely that Flush() does something without having rawSpans.

codeboten:
hey @szuecs, I created a PR to your fork https://github.com/szuecs/lightstep-tracer-go/pull/1. The change fixes one issue which was breaking some of the tests where `parts` was never set if `heuristicByteSize` was not greater than `tracer.opts.GRPCMaxCallSendMsgSizeBytes`

I've also restored the behaviour to go through `Flush` for now even when there are no rawSpans. Will need to do a bit more digging on our end to see if removing this impacts anything else.

szuecs:
Thanks @codeboten !

jmacd (approved):

PR #279, opened 2021-04-14
Propagate sampling decisions for binary carrier

lorenmh:
_Sorry for prematurely opening prior PR (#278), meant to test this in a private forked repo before opening the PR._

The binary carrier was not propagating sampling decisions, PR fixes this so the sampling decision is correctly propagated. The sampling decision wasn't being used after decoding from the wire format; PR simply uses the wire format's sampling decision when decoding.

### DIFF
--- a/tracer_transport_test.go
+++ b/tracer_transport_test.go
@@ -256,6 +256,12 @@ var _ = Describe("Tracer Transports", func() {
 					TraceID: 456000000000,
 					Baggage: map[string]string{"a": "1", "b": "2", "c": "3"},
 				}
+				var testContextSampledFalse = SpanContext{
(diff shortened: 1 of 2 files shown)


tracer_transport_test.go
```diff
@@ -306,6 +312,15 @@ var _ = Describe("Tracer Transports", func() {
 						err = tracer.Inject(nil, opentracing.Binary, carrierBytes)
 						Expect(err).To(HaveOccurred())
 					})
+					It("Should propagate sampled", func() {
```
lorenmh, line 315: this test fails without the changes in `propagation_binary.go`

codeboten (approved):

PR #280, opened 2021-04-19
Update gopsutil to v3 for macOS/arm64 (Apple M1) compatibility 

alexfjw:
Hi, I've updated gopsutil to v3 so macOS/arm64 devices can build this project locally. 

See https://github.com/shirou/gopsutil/issues/1000 regarding gopsutil-v2 incompatibility with macOS/arm64. 
There are similar updates done on other open source repos, eg. https://github.com/kubernetes/minikube/pull/10115

Tested locally with `go version go1.16 darwin/arm64`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Pending Release](https://github.com/lightstep/lightstep-tracer-go/compare/v0.24.0...HEAD)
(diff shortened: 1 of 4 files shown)


whopper:
👍 this is definitely needed!

codeboten (approved):
Thanks for the contrib @alexfjw!

